PR #1, opened 2021-02-20
Add documentation, comments and some code improvements 

dlsf:


src/main/java/com/iridium/iridiumskyblock/Item.java
```diff
@@ -4,57 +4,115 @@
...
-    public String title;
+    public String displayName;
     public String headData;
     public String headOwner;
     public List<String> lore;
     public Integer slot;
 
+    /**
+     * Default, empty constructor.
+     * No one knows why it exists in the first place.
```
PeachesMLG, line 23: Because the Jackson serializer needs a no args constructor for it to load stutff

src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
```diff
@@ -6,74 +6,123 @@
...
-                }
-            }
-        } else {
-            if (cs instanceof Player) {
-                Player p = (Player) cs;
+    public boolean onCommand(@NotNull CommandSender commandSender, org.bukkit.command.@NotNull Command cmd, @NotNull String label, String[] args) {
+        if (args.length == 0) {
+            if (commandSender instanceof Player) {
+                Player p = (Player) commandSender;
+                // TODO: Send message
```
PeachesMLG, line 75: Why was this deleted?

src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
```diff
@@ -6,74 +6,123 @@
...
+     * @param command The command which should be registered
+     */
     public void registerCommand(Command command) {
         commands.add(command);
     }
 
-    public void unRegisterCommand(Command command) {
+    /**
+     * Unregisters a single command in the command system.
+     * @param command The command which should be registered
```
PeachesMLG, line 53: unregisted :P

src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
PeachesMLG: Nevermind :eyes:

src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
```diff
@@ -6,74 +6,123 @@
...
+                Player p = (Player) commandSender;
+                // TODO: Send message
                 return true;
             }
         }
-        cs.sendMessage(StringUtils.color(iridiumSkyblock.getMessages().unknownCommand
+
+        for (Command command : commands) {
+            // We don't want to execute other commands or ones that are disabled
+            if (!(command.aliases.contains(args[0]) && command.enabled)) {
```
PeachesMLG, line 82: this shouldnt be casesensitive

src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
```diff
@@ -6,74 +6,123 @@
...
+            }
+
+            if (command.player && !(commandSender instanceof Player)) {
+                // Must be a player
+                commandSender.sendMessage(StringUtils.color(iridiumSkyblock.getMessages().mustBeAPlayer
+                        .replace("%prefix%", iridiumSkyblock.getConfiguration().prefix)));
+                return false;
+            }
+
+            if (!((commandSender.hasPermission(command.permission) || command.permission
```
PeachesMLG, line 93: if the command.permission is empty then there is no permission required. so it should be && !command.permission.equals("")

src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
```diff
@@ -6,74 +6,123 @@
...
+            }
+
+            command.execute(commandSender, args);
+            return true;
+        }
+        // Unknown command message
+        commandSender.sendMessage(StringUtils.color(iridiumSkyblock.getMessages().unknownCommand
                 .replace("%prefix%", iridiumSkyblock.getConfiguration().prefix)));
-        return true;
+        return false;
```
PeachesMLG, line 108: Why return false?

src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
```diff
@@ -1,9 +1,15 @@
 package com.iridium.iridiumskyblock.configs;
 
+/**
+ * The main configuration of IridiumSkyblock.
+ * Is persisted automatically on plugin shutdown and reload.
```
PeachesMLG, line 5: this isnt correct, it doesnt persist on shutdown because if somone made changes to the config then did /rl it would save the config (overwriting their changes) and this would annoy people

PeachesMLG: Also wrong for Messages & SQL

src/main/java/com/iridium/iridiumskyblock/utils/PlayerUtils.java
```diff
@@ -13,41 +22,28 @@ private static int getExpAtLevel(final int level) {
...
-
-        while (currentLevel > 0) {
-            currentLevel--;
-            exp += getExpAtLevel(currentLevel);
-        }
-        if (exp < 0) {
-            exp = Integer.MAX_VALUE;
-        }
-        return exp;
+        return player.getTotalExperience();
```
PeachesMLG, line 33: I cant remember the exact reason but I did this on purpose. I belive that player.getTotalExperiance was sometimes wrong? I cant quite recall exactly

src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
dlsf: Then it was not case sensitive in the first place, I just applied the early return patern

src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
dlsf: Because you should return false when the command is not valid... It doesn't really matter tho

https://hub.spigotmc.org/javadocs/spigot/org/bukkit/command/CommandExecutor.html#onCommand(org.bukkit.command.CommandSender,org.bukkit.command.Command,java.lang.String,java.lang.String%5B%5D)

src/main/java/com/iridium/iridiumskyblock/utils/PlayerUtils.java
dlsf: I guess that could be the case (https://bukkit.org/threads/p-gettotalexperience-acting-weird.407546/#post-3331133), but I think it works correctly on newer versions of Spigot. It depends if you want to keep 1.8 - 1.12 support

src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
dlsf: I just inverted the whole statement, it should work

src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
PeachesMLG: oh it wasnt... well it was supposed to be lol. Ill take a look at that later

src/main/java/com/iridium/iridiumskyblock/utils/PlayerUtils.java
PeachesMLG: Ah ye, that sounds about right. Keep it has it was before. Im going to try and keep 1.8-1.12 depending on how much of a headache it gives me

PR #4, opened 2021-02-22
Add more documentation and clean up code

dlsf:


src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
```diff
@@ -11,7 +11,11 @@
 import java.util.HashMap;
 import java.util.concurrent.CompletableFuture;
 
+/**
+ * Class which handles schematics und pastes them.
```
PeachesMLG, line 15: and*

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
dlsf: Just wrote the German and

PR #11, opened 2021-02-27
Some issue fixes

MertUnverdi:


src/main/java/com/iridium/iridiumskyblock/database/Island.java
```diff
@@ -115,6 +106,15 @@ public void setHome(@NotNull Location location) {
         this.home = homeLocation.getX() + "," + homeLocation.getY() + "," + homeLocation.getZ() + "," + homeLocation.getPitch() + "," + homeLocation.getYaw();
     }
 
+    /**
+     * Milliseconds of date this island was created
+     *
+     * @return A LocalDateTime of this island was created
+     */
+    public LocalDateTime getCreateTime() {
```
MertUnverdi, line 114: I just wrapped the setHome and getHome methods at here

src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
```diff
@@ -23,6 +23,7 @@
     public String invitedPlayer = "%prefix% &7You have invited %player% to join your island";
     public String userInvitedPlayer = "%prefix% &7%inviter% has invited %player% to join your island";
     public String youHaveBeenInvited = "%prefix% &7%inviter% has invited you to join their island";
+    public String alreadyInYourIsland = "%prefix% &7This player is already in your island team";
```
PeachesMLG, line 26: change to
     public String alreadyInYourIsland = "%prefix% &7This player is already in your island .";

src/main/java/com/iridium/iridiumskyblock/gui/MembersGUI.java
```diff
@@ -31,9 +31,9 @@ public MembersGUI(@NotNull Island island) {
     public void onInventoryClick(InventoryClickEvent event) {
         if (members.containsKey(event.getSlot())) {
             User user = members.get(event.getSlot());
-            if (event.getClick().equals(ClickType.RIGHT)) {
+            if (event.getClick().equals(ClickType.LEFT)) {
```
PeachesMLG, line 34: I agree left seems left seems like it should demote and right should promote, but if your going to change this you also need to change the lore in inventories.json

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/gui/MembersGUI.java
MertUnverdi: i changed it cuz that already promote = right   demot = left in inventories.(YML)

MertUnverdi: and should add else if cause users can be click with shift or middle click

PeachesMLG (approved):

src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
```diff
@@ -23,6 +23,7 @@
     public String invitedPlayer = "%prefix% &7You have invited %player% to join your island";
     public String userInvitedPlayer = "%prefix% &7%inviter% has invited %player% to join your island";
     public String youHaveBeenInvited = "%prefix% &7%inviter% has invited you to join their island";
+    public String alreadyInYourIsland = "%prefix% &7This player is already in your island.";
```
dlsf, line 26: "is already a member of your island" probably would be better

dlsf (changes requested):

PR #12, opened 2021-02-27
Add Gradle 

dlsf:


build.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+    pdm("com.j256.ormlite:ormlite-jdbc:5.3")
+    pdm("com.fasterxml.jackson.core:jackson-databind:2.12.1")
+    pdm("com.fasterxml.jackson.core:jackson-core:2.12.1")
+    pdm("com.fasterxml.jackson.core:jackson-annotations:2.12.1")
+    pdm("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.12.1")
+
+    // Other dependencies that are not required or already available at runtime
+    compileOnly("org.spigotmc:spigot-api:1.16.5-R0.1-SNAPSHOT")
+    compileOnly("org.spigotmc:spigot:1.16.4")
+    compileOnly("org.yaml:snakeyaml:1.27")
```
PeachesMLG, line 37: you need to shade this in (or download it with pdm) otherwise it wont work on 1.8

PeachesMLG: (Also need to relocate it)

PeachesMLG (changes requested):

build.gradle.kts
dlsf: I'll test if this is an issue

BomBardyGamer (approved):

build.gradle.kts
PeachesMLG: Were  talking about snakeyaml  here  not the  spigot api xD

src/main/java/com/iridium/iridiumskyblock/IridiumSkyblock.java
```diff
@@ -122,7 +122,13 @@ public void onEnable() {
     private void loadDependencies() {
         PluginDependencyManager dependencyManager = SpigotDependencyManager.of(this);
         CompletableFuture<Void> loadAllDependencies = dependencyManager.loadAllDependencies();
```
PeachesMLG, line 124: So this is being done async. What happens if we try to access a dependency while its still being downloaded? e.g. when we serialize the configs and stuff

PeachesMLG: Although I just realised that its doing this onLoad not onEnable, but what happens if the server is down or its lagging alot and onEnable gets called before its finished?

BomBardyGamer: Yeah this needs to be called blocking anyway

BomBardyGamer (approved):

src/main/java/com/iridium/iridiumskyblock/IridiumSkyblock.java
dlsf: It is blocking?

BomBardyGamer: Yeah it is now

BomBardyGamer: Also, why don't you just call `.join()` instead of calling `.get()` which throws exceptions

BomBardyGamer: Ah actually I see

PeachesMLG:
Need to find a way to relocate snake yaml before this can be merged to solve issues with 1.8 clients

BomBardyGamer:
Or maybe we should just completely drop support for 1.8. I mean, it is what, 7 years old now?

dlsf:
I agree

PR #25, opened 2021-03-20
Added Island Bank

PeachesMLG:


src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -233,6 +234,24 @@ public boolean getIslandPermission(@NotNull Island island, @NotNull User user, @
...
 
+    /**
+     * Gets an Island's bank from BankItem
+     *
+     * @param island   The specified Island
+     * @param bankItem The BankItem we are getting
+     * @return the IslandBank
+     */
+    public IslandBank getIslandBank(@NotNull Island island, @NotNull BankItem bankItem) {
+        Optional<IslandBank> optionalIslandBank = IridiumSkyblock.getInstance().getDatabaseManager().getIslandBankList().stream().filter(isBank -> isBank.getIsland() == island.getId() && isBank.getBankItem().equalsIgnoreCase(bankItem.getName())).findFirst();
```
dlsf, line 245: isBank sounds like an accessor for a boolean, islandBank or just bank would be better choices

src/main/resources/plugin.yml
```diff
@@ -4,6 +4,8 @@ description: Skyblock Redifined
 main: com.iridium.iridiumskyblock.IridiumSkyblock
 api-version: '1.13'
 author: Peaches_MLG
+depend:
+  - Vault
```
dlsf, line 8: Either do a list in this style or in the style of the alias list

dlsf (changes requested):

PR #30, opened 2021-04-03
Missions

PeachesMLG:
Added Missions
(Nearly finished)


src/main/java/com/iridium/iridiumskyblock/database/Island.java
```diff
@@ -71,6 +74,16 @@ public Island(@NotNull String name, @NotNull Schematics.SchematicConfig schemati
...
+        this.experience = 0;
+    }
+
+    /**
+     * Gets the island's level
+     *
+     * @return The islands level
+     */
+    public int getLevel() {
+        return (int) Math.cbrt(experience + 1);
```
dlsf, line 86: This is a pretty bad way to calculate the level IMO

src/main/java/com/iridium/iridiumskyblock/Reward.java
```diff
@@ -0,0 +1,29 @@
...
+import org.bukkit.entity.Player;
+
+import java.util.List;
+
+@NoArgsConstructor
+public class Reward {
+    public Item item;
+    public List<String> commands;
+    public int islandExperience;
+    public XSound sound;
```
dlsf, line 16: Why do you have a sound for every reward if you only use the one of the mission?

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -378,6 +416,68 @@ private void recalculateIsland(@NotNull Island island, @NotNull List<ChunkSnapsh
...
     }
 
+    /**
+     * Increments a mission's data based on requirements
+     *
+     * @param island      The island
+     * @param missionData The mission data e.g. BREAK:COBBLESTONE
+     * @param increment   The amount we are incrementing by
+     */
+    public void incrementMission(@NotNull Island island, @NotNull String missionData, int increment) {
```
dlsf, line 426: This method is a mess, split it up and add comments or something like that

dlsf (changes requested):
I've seen worse things

src/main/java/com/iridium/iridiumskyblock/database/Island.java
dlsf: There are multiple reasons for this.
First off, Math.cbrt() is very slow by a large margin even in comparison to other functions like Math.sqrt() which are already considered to be too slow: https://replit.com/@Dasunterstrich/InternationalSeagreenRom
Next, the graph for this function is simply not good for levels. The first three levels only need 99 XP, while level 10 already requires 1000, which in my opinion is too steep / a too big difference.

src/main/java/com/iridium/iridiumskyblock/Reward.java
PeachesMLG: Rewards havent been implemented yet, alot of people wanted commands as rewards so to do this ima have a /is rewards gui the sound is going to be the sound that plays where you collect the rewards from that GUI

src/main/java/com/iridium/iridiumskyblock/database/Island.java
PeachesMLG: ye I didnt really know what to do for levels feel free to replace it with another function at a different time, but for now ima just leave it for testing purposes

PeachesMLG: Also its not that big of a deal according to your code cube rooting 1000000 times only takes 130ms, shouldnt be that big of a deal

PR #40, opened 2021-04-10
Changed island partition on visit gui

MertUnverdi:
Used Lists#partition instead of Stream#skip thing

** pages is now starts at 0 **


dlsf:
I don't really know how to feel about this. It's certainly a little bit shorter, but also messier (line 29, 30, 38 & 39 took me 5 minutes to fully understand + find Javadocs for). Also, I'm pretty sure that the skip and limit thingy is faster because it influences the number of entries in the Stream that have to be processed.

PeachesMLG:
Not sure how I feel about 53 and 54, also what if there are no islands, will calling
`        this.islands = Lists.partition(IridiumSkyblock.getInstance().getDatabaseManager().getIslandList().stream().filter(Island::isVisitable)
                .collect(Collectors.toList()), elementsPerPage).get(page);`
return null or an empty list? other than that I like it

PR #44, opened 2021-04-15
Added PlaceholderBuilder

PeachesMLG:
Added PlaceholderBuilder to provide more continuity between placeholders

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/PlaceholderBuilder.java
+++ b/src/main/java/com/iridium/iridiumskyblock/PlaceholderBuilder.java
@@ -1,0 +1,41 @@
+package com.iridium.iridiumskyblock;
+
+import com.iridium.iridiumskyblock.database.Island;
+import com.iridium.iridiumskyblock.database.IslandBlocks;
+import com.iridium.iridiumskyblock.database.User;
+import com.iridium.iridiumskyblock.utils.Placeholder;
+
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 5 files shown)


src/main/java/com/iridium/iridiumskyblock/PlaceholderBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.List;
+
+public class PlaceholderBuilder {
+    private final List<Placeholder> placeholderList = new ArrayList<>();
+
+    public PlaceholderBuilder() {
+        placeholderList.add(new Placeholder("prefix", IridiumSkyblock.getInstance().getConfiguration().prefix));
```
dlsf, line 16: You should probably write a method to do this... Much duplicate code at the moment

dlsf (changes requested):

src/main/java/com/iridium/iridiumskyblock/PlaceholderBuilder.java
PeachesMLG: a method to do what? I am confusion

dlsf: To register a placeholder replacement

PeachesMLG: No cuz if its an object we can parse it to ItemStacks and other sutff, anyways ima merge this. if u think its shit do a pr
:heart:

PR #48, opened 2021-04-25
Improve static factory methods of enums

dlsf:


PeachesMLG:
Main issue is valueof throws an exception instead of returning null

dlsf:
How do we handle null values at the moment? Wouldn't it be better to just return a fallback option?

PeachesMLG:
no, if its null we just send the player a message like "Thats not a thing idot"

dlsf:
I see, we would have to add a try-catch with this approach. What option do you prefer?

PeachesMLG:
I feel like how we do it now is fine where we just return null

PR #52, opened 2021-04-26
Clean up IslandManager

PeachesMLG:
Cleaning up IslandManager and remove duplicate Code

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/managers/TableManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/TableManager.java
@@ -1,0 +1,113 @@
+package com.iridium.iridiumskyblock.managers;
+
+import com.j256.ormlite.dao.Dao;
+import com.j256.ormlite.dao.DaoManager;
+import com.j256.ormlite.support.ConnectionSource;
+import com.j256.ormlite.support.DatabaseConnection;
+import com.j256.ormlite.table.TableUtils;
+
+import java.sql.SQLException;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * @param <T> The Table Class
+ * @param <S> The Table Primary Id Class
+ */
+public class TableManager<T, S> {
+    private final List<T> list;
+    private final Dao<T, S> dao;
+
+    private final boolean autoCommit;
+    private final ConnectionSource connectionSource;
+
+    public TableManager(ConnectionSource connectionSource, Class<T> clazz, boolean autoCommit) throws SQLException {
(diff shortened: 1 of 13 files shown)


src/main/java/com/iridium/iridiumskyblock/managers/TableManager.java
```diff
@@ -0,0 +1,110 @@
...
+import com.j256.ormlite.support.ConnectionSource;
+import com.j256.ormlite.support.DatabaseConnection;
+import com.j256.ormlite.table.TableUtils;
+
+import java.sql.SQLException;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * @param <T> The Table Class
```
dlsf, line 14: Add a short description when you add Javadocs

src/main/java/com/iridium/iridiumskyblock/managers/TableManager.java
```diff
@@ -0,0 +1,110 @@
...
+        this.dao.setAutoCommit(getDatabaseConnection(), autoCommit);
+        this.list = dao.queryForAll();
+    }
+
+    /**
+     * Saves everything to the Database
+     */
+    public void save() {
+        try {
+            for (T t : list) {
```
dlsf, line 38: list.forEach(dao::createOrUpdate) is way cooler

dlsf (changes requested):

src/main/java/com/iridium/iridiumskyblock/managers/TableManager.java
```diff
@@ -0,0 +1,110 @@
...
+import java.sql.SQLException;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * @param <T> The Table Class
+ * @param <S> The Table Primary Id Class
+ */
+public class TableManager<T, S> {
+    private final List<T> list;
```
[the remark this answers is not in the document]
dlsf: Maybe not, but list is too generic

src/main/java/com/iridium/iridiumskyblock/managers/TableManager.java
PeachesMLG: k but u wrong tho.

Cuz then if I do that the try/catch wont work since dao::createOrUpdate is its own function so it will be more like 
```
            list.forEach(t -> {
                try {
                    dao.createOrUpdate(t);
                } catch (SQLException throwables) {
                    throwables.printStackTrace();
                }
            });
``` 
which is bad and dum

PR #61, opened 2021-05-10
Add Ocean generator, Improve Skyblock generator

dlsf:


plugin/src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
```diff
@@ -36,4 +42,23 @@
...
+    /**
+     * Settings for the {@link org.bukkit.generator.ChunkGenerator} of IridiumSkyblock.
+     * Allows fine-tuning of the {@link com.iridium.iridiumskyblock.generators.OceanGenerator}.
+     */
+    @AllArgsConstructor
+    @NoArgsConstructor
+    public static class GeneratorSettings {
+
+        public GeneratorType generatorType = GeneratorType.SKYBLOCK;
+        public XBiome defaultOverworldBiome = XBiome.PLAINS;
```
PeachesMLG, line 54: These arnt needed since we will be defining the biome with the island schematic

PeachesMLG (changes requested):

PR #80, opened 2021-05-27
Potential fix of java.lang.IllegalArgumentException

Enter25565:
Potential fix of exception java.lang.IllegalArgumentException: y out of range (expected 0-255, got 256).
Full log: [https://pastebin.com/FyTqu8bP](https://pastebin.com/FyTqu8bP)
My fix ensures, that `maxy` is lower or equal to 255, by using `Math.min(...)` method.
Please test it before accepting.

### DIFF
--- a/plugin/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
+++ b/plugin/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
@@ -604,7 +604,7 @@ public class IslandManager {
                             for (int x = 0; x < 16; x++) {
                                 for (int z = 0; z < 16; z++) {
                                     if (island.isInIsland(x + (chunk.getX() * 16), z + (chunk.getZ() * 16))) {
-                                        final int maxy = chunk.getHighestBlockYAt(x, z);
+                                        final int maxy = Math.min(255, chunk.getHighestBlockYAt(x, z));
                                         for (int y = 0; y <= maxy; y++) {
                                             XMaterial material = IridiumSkyblock.getInstance().getMultiversion().getMaterialAtPosition(chunk, x, y, z);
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -604,7 +604,7 @@ private void recalculateIsland(@NotNull Island island, @NotNull List<Chunk> chun
                             for (int x = 0; x < 16; x++) {
                                 for (int z = 0; z < 16; z++) {
                                     if (island.isInIsland(x + (chunk.getX() * 16), z + (chunk.getZ() * 16))) {
-                                        final int maxy = chunk.getHighestBlockYAt(x, z);
+                                        final int maxy = Math.min(255, chunk.getHighestBlockYAt(x, z));
                                         for (int y = 0; y <= maxy; y++) {
```
dlsf, line 608: Maybe remove the `=` here instead of using the Math.min() method

dlsf (changes requested):
This suggestion should work too, however I'm not sure if this issue is not already fixed so we'll just wait for @PeachesMLG

plugin/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
PeachesMLG: This wont work cuz spigot is dum and slightly changed this method at one point, it used to be the highest block, now its the block above the highest block

dlsf: And removing the = would only count until one block below so it would work again 🤔

PR #82, opened 2021-05-28
Fix ocean generator regeneration issues

dlsf:


plugin/src/main/java/com/iridium/iridiumskyblock/generators/OceanGenerator.java
```diff
@@ -61,16 +61,78 @@
...
+        XMaterial bottomMaterial = IridiumSkyblock.getInstance().getConfiguration().generatorSettings.oceanFloorBottomMaterial;
+        XMaterial topMaterial = IridiumSkyblock.getInstance().getConfiguration().generatorSettings.oceanFloorTopMaterial;
+        int waterHeight = IridiumSkyblock.getInstance().getConfiguration().generatorSettings.waterHeight;
+        int maxOceanFloorLevel = IridiumSkyblock.getInstance().getConfiguration().generatorSettings.maxOceanFloorLevel;
+        int minOceanFloorLevel = IridiumSkyblock.getInstance().getConfiguration().generatorSettings.minOceanFloorLevel;
+
+        int currentFloorHeight = (int) ((generator.noise(x, z, 1.5D, 0.5D, true) + 1) * (maxOceanFloorLevel - minOceanFloorLevel) + minOceanFloorLevel);
+
+        // Generate layer of bedrock
+        IridiumSkyblock.getInstance().getNms().setBlockFast(
```
PeachesMLG, line 87: This wont work properly, We need to delete the blocks that should be air, also for performance we should check if the block were setting isnt already set

PeachesMLG (changes requested):

PR #84, opened 2021-05-28
Add number abbreviations

dlsf:


dlsf:
Also adds a few more placeholders

PR #91, opened 2021-05-29

### LINKED ISSUE #87: SpawnerSpawnEvent is called once per entity spawn

Nubebuster:
closes #87 Fixed the issue by putting the spawner in a lock list and removing it from the lock the next tick, when the spawn delay is halved.


Nubebuster:
I am having trouble with squashing commits today haha

PR #92, opened 2021-05-29

### LINKED ISSUE #87: SpawnerSpawnEvent is called once per entity spawn

Nubebuster:
closes #87 Fixes the issue by putting the spawner in a lock list and removing it from the lock the next tick, when the spawn delay is halved.

### DIFF
--- a/plugin/src/main/java/com/iridium/iridiumskyblock/listeners/SpawnerSpawnListener.java
+++ b/plugin/src/main/java/com/iridium/iridiumskyblock/listeners/SpawnerSpawnListener.java
@@ -4,13 +4,22 @@ import com.iridium.iridiumskyblock.IridiumSkyblock;
 import com.iridium.iridiumskyblock.database.Island;
 import com.iridium.iridiumskyblock.database.IslandBooster;
 import org.bukkit.Bukkit;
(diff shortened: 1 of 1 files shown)


dlsf (approved):

plugin/src/main/java/com/iridium/iridiumskyblock/listeners/SpawnerSpawnListener.java
```diff
@@ -4,23 +4,37 @@
...
+
     @EventHandler
     public void onCreatureSpawn(SpawnerSpawnEvent event) {
         Optional<Island> island = IridiumSkyblock.getInstance().getIslandManager().getIslandViaLocation(event.getLocation());
         if (island.isPresent()) {
             IslandBooster islandBooster = IridiumSkyblock.getInstance().getIslandManager().getIslandBooster(island.get(), "spawner");
             if (islandBooster.isActive()) {
+                CreatureSpawner spawner = event.getSpawner();
+                if(lock.contains(spawner))
+                    return;
```
dlsf, line 32: We do that in one line

PeachesMLG:
This seems very hack, im currently looking to see if theres a better way to do this

PR #96, opened 2021-05-29
Replace reflection for multi-version support with safe alternative

dlsf:
Uses an enum instead of reflections to access the multi-version support and fixes an issue with the V1_8_R3 multi-version class having a wrong name


dlsf:
I renamed the multiversion and nms classes so we don't have to use the whole package names in the enum.

PR #103, opened 2021-05-31
Fix ClassNotFoundException for MultiVersion code on HotSpot VM

dlsf:


PeachesMLG:
Looks good, just add a comment explaining why we are doing like this so somone doesnt try and do a pr and revert this

PR #117, opened 2021-06-03
Member Upgrade and new Placeholders

MertUnverdi:
Added Member Upgrade

Add Island Upgrade placeholders

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/PlaceholderBuilder.java
+++ b/src/main/java/com/iridium/iridiumskyblock/PlaceholderBuilder.java
@@ -21,6 +21,7 @@ public class PlaceholderBuilder {
         placeholderList.add(new Placeholder("island_value", String.valueOf(island.getValue())));
         placeholderList.add(new Placeholder("island_rank", String.valueOf(island.getRank())));
         placeholderList.add(new Placeholder("island_level", String.valueOf(island.getLevel())));
+        placeholderList.add(new Placeholder("island_members", String.valueOf(island.getMembers().size())));
         placeholderList.add(new Placeholder("island_create", island.getCreateTime().format(DateTimeFormatter.ofPattern(IridiumSkyblock.getInstance().getConfiguration().dateTimeFormat))));
 
         IridiumSkyblock.getInstance().getBlockValues().blockValues.keySet().stream().map(material -> new Placeholder(material.name() + "_AMOUNT", String.valueOf(IridiumSkyblock.getInstance().getIslandManager().getIslandBlock(island, material).getAmount()))).forEach(placeholderList::add);
--- a/src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
+++ b/src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
@@ -36,7 +36,7 @@ public class Messages {
     public String youHaveBeenInvited = "%prefix% &7%inviter% has invited you to join their island.";
     public String alreadyInYourIsland = "%prefix% &7This player is already a member of your Island.";
     public String inviteRevoked = "%prefix% &7Island invite for %player% has been revoked.";
-    public String islandTeamFull = "%prefix% &7You cannot invite more members to your island.";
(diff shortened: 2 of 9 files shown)


src/main/java/com/iridium/iridiumskyblock/upgrades/SizeUpgrade.java
```diff
@@ -19,6 +19,6 @@ public SizeUpgrade(int money, int crystals, int size) {
     @JsonIgnore
     @Override
     public List<Placeholder> getPlaceholders() {
-        return Collections.singletonList(new Placeholder("size", String.valueOf(size)));
+        return Arrays.asList(new Placeholder("dimensions", String.valueOf(size)), new Placeholder("size", String.valueOf(size)));
```
PeachesMLG, line 22: Whats the point of this?

MertUnverdi: may old users still remember it as "dimension" instead "size"
So i've just added an alternative

PeachesMLG: I mean I dont see this being a massive issue since its default as size, but like :shrug:

src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
```diff
@@ -36,6 +36,7 @@
     public String youHaveBeenInvited = "%prefix% &7%inviter% has invited you to join their island.";
     public String alreadyInYourIsland = "%prefix% &7This player is already a member of your Island.";
     public String inviteRevoked = "%prefix% &7Island invite for %player% has been revoked.";
+    public String islandTeamFull = "%prefix% &7Island maximum member count reached";
```
dlsf, line 39: Sounds a little bit odd, mostly because there is no verb.

Maybe "The maximum amount of members has already been reached"?

dlsf (changes requested):

src/main/java/com/iridium/iridiumskyblock/PlaceholderBuilder.java
```diff
@@ -21,6 +21,7 @@ public PlaceholderBuilder applyIslandPlaceholders(Island island) {
         placeholderList.add(new Placeholder("island_value", String.valueOf(island.getValue())));
         placeholderList.add(new Placeholder("island_rank", String.valueOf(island.getRank())));
         placeholderList.add(new Placeholder("island_level", String.valueOf(island.getLevel())));
+        placeholderList.add(new Placeholder("island_members", String.valueOf(island.getMembers().size())));
```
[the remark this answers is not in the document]
MertUnverdi: is already on placeholders i tought here just for guis

src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
MertUnverdi: my bad english..

src/main/java/com/iridium/iridiumskyblock/PlaceholderBuilder.java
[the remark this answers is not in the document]
dlsf: Might still be wanted in GUIs

src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
MertUnverdi: Oh wait this sends on player tried join a full island too
May it should om another message what your things ?

dlsf: Yes you should

dlsf: It may only be sent to the player who can't join

PeachesMLG: How about something like "This island is currently full" or "The island member count has been reached" or smthn

PR #119, opened 2021-06-04
Member Upgrade and new placeholders

MertUnverdi:


PeachesMLG:
looks good to me, but I usually miss stuff, so ima wait until das reviews

dlsf (approved):

PR #124, opened 2021-06-05
Add config to Respawn on island

MertUnverdi:


src/main/java/com/iridium/iridiumskyblock/listeners/PlayerRespawnListener.java
```diff
@@ -0,0 +1,18 @@
...
+import org.bukkit.event.Listener;
+import org.bukkit.event.player.PlayerRespawnEvent;
+
+public class PlayerRespawnListener implements Listener {
+
+    @EventHandler
+    public void onRespawn(PlayerRespawnEvent event) {
+        if (IridiumSkyblock.getInstance().getConfiguration().respawnOnIsland) {
+            User user = IridiumSkyblock.getInstance().getUserManager().getUser(event.getPlayer());
+            user.getIsland().ifPresent(island -> event.setRespawnLocation(island.getHome()));
```
dlsf, line 15: I think you should check if the current respawn location is already on the island because of beds, respawn anchors etc

PeachesMLG, line 15: Replace with
```
LocationUtils.getSafeLocation(island.getHome(), island)
```

PR #125, opened 2021-06-05
Add ability to disable the island shop

MertUnverdi:


dlsf:
Useless. They can disable the command anyways

PR #127, opened 2021-06-05
Added More Bank SubCommands

PeachesMLG:


src/main/java/com/iridium/iridiumskyblock/commands/BankGive.java
```diff
@@ -0,0 +1,70 @@
+package com.iridium.iridiumskyblock.commands;
```
dlsf, line 1: I guess it's not a good idea to have these in the normal commands package because it'd just be confusing. They behave differently to normal commands. Maybe put them in a subcommand or bank sub-package?

src/main/java/com/iridium/iridiumskyblock/commands/BankCommand.java
```diff
@@ -2,31 +2,39 @@
...
+import java.util.stream.Stream;
 
 /**
  * Command which opens the Island bank GUI.
  */
 public class BankCommand extends Command {
 
+    public BankGive bankGive;
+    public BankSet bankSet;
+    public BankRemove bankRemove;
```
dlsf, line 28: You can't configure these commands, can you?

PeachesMLG: yep you can, the bank command gets serialized, which includes all objects inside it

PeachesMLG: Gets created like this https://pastebin.com/p12PFBqp

PR #135, opened 2021-06-06
Added Island Levelup Rewards

PeachesMLG:


dlsf (changes requested):
Looks pretty good otherwise. You could replace the for loops with stream I guess but that's just optional

src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
```diff
@@ -62,6 +69,32 @@
...
+                    "&b&l[!] &bLeft click to redeem"
+            )), Collections.emptyList(), 0, 5, 1000, 0, XSound.ENTITY_PLAYER_LEVELUP))
+
+            .put(5, new Reward(new Item(XMaterial.EXPERIENCE_BOTTLE, 1, "&b&lLevel %island_level% Reward", Arrays.asList(
+                    "&7Island Level %island_level% Rewards:",
+                    "&b&l* &b15 Island Crystals",
+                    "&b&l* &b10000 Island Money",
+                    "",
+                    "&b&l[!] &bLeft click to redeem"
+            )), Collections.emptyList(), 0, 15, 10000, 0, XSound.ENTITY_PLAYER_LEVELUP))
```
dlsf, line 95: I think that these default rewards are too high, you get to these levels pretty quickly if I recall correctly. 

I'd say 200 / 5 for level 1 and 2000 / 15 for level 5

dlsf: Considering the shop prices too. You can't give a new player enough money to buy some advanced stuff

PR #143, opened 2021-06-10
Make rank names configurable

MertUnverdi:


PeachesMLG:
I think there was a suggestion to make the items configurable, how about doing that instead?

dlsf:
I agree, that'd probably be better

MertUnverdi:
Ya its better, i'll do when at home thanks for reviews

dlsf:
You should revert the changes from your previous attempt

MertUnverdi:
Make new pr (i couldn't revert previous commits)

PR #153, opened 2021-06-14

### LINKED ISSUE #87: SpawnerSpawnEvent is called once per entity spawn

Nubebuster:
closes #87 Fixes the issue by putting the spawner in a lock list and removing it from the lock the next tick, when the spawn delay is halved. Note: Slighly improved version compared to what I comitted before.


dlsf:
Sorry for the wait. I'm currently working on an advanced Cooldown system which can also be used for this issue so this is no longer needed.

Thanks for you contribution anyways!

PR #154, opened 2021-06-14
Optimize BlockFormListener and allow ores to be generated from solid …

Nubebuster:
…stone too (which is formed from a water source block).

This patch changes the BlockFormListener from parsing ore frequencies every time a block is formed to parsing at the point of creating the blockformlistener instance. This will put a tiny amount of data into memory but will save a lot of cpu time when the server is running. This also parses the XMaterial objects to Material objects which also saves a bit of cpu time.

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
+++ b/src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
@@ -1,9 +1,9 @@
 package com.iridium.iridiumskyblock.listeners;
 
+import com.google.gson.internal.LinkedTreeMap;
 import com.iridium.iridiumcore.dependencies.xseries.XMaterial;
 import com.iridium.iridiumskyblock.IridiumSkyblock;
 import com.iridium.iridiumskyblock.database.Island;
-import com.iridium.iridiumskyblock.database.IslandUpgrade;
 import com.iridium.iridiumskyblock.upgrades.OresUpgrade;
 import org.bukkit.Material;
 import org.bukkit.World;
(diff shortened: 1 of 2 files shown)


src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
```diff
@@ -1,42 +1,78 @@
...
-                for (Map.Entry<XMaterial, Integer> entries : ores.entrySet()) {
-                    for (int i = 0; i < entries.getValue(); i++) {
-                        materials.add(entries.getKey());
-                    }
-                }
-
-                Material material = materials.get(random.nextInt(materials.size())).parseMaterial();
+                Material[] materials = upgradesParsed.get(IridiumSkyblock.getInstance().getIslandManager().getIslandUpgrade(island.get(), "generator").getLevel())[newMaterial.equals(XMaterial.BASALT) ? 1 : 0];
+                Material material = materials[random.nextInt(materials.length)];
+                if (material == Material.COBBLESTONE && event.getBlock().getType() == Material.STONE)
```
PeachesMLG, line 72: I belive this event gets called before it actually sets the block so you need to change event.getBlock().getType() to newMaterial

src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
```diff
@@ -1,42 +1,78 @@
...
+            oresArrays[1] = new Material[netherOreFrequencySum];
+            int indexNether = 0;
+            for (Map.Entry<XMaterial, Integer> entries : upgrades.getValue().netherOres.entrySet()) {
+                Material material = entries.getKey().parseMaterial();
+                for (int i = 0; i < entries.getValue(); i++) {
+                    oresArrays[1][indexNether++] = material;
+                }
+            }
+            upgradesParsed.put(upgrades.getKey(), oresArrays);
+        }
```
dlsf, line 60: This looks really messy. Using a two-dimensional array is weird, why not another array? Also, what about /is reload?

src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
```diff
@@ -1,42 +1,78 @@
...
-                Random random = new Random();
-
-                for (Map.Entry<XMaterial, Integer> entries : ores.entrySet()) {
-                    for (int i = 0; i < entries.getValue(); i++) {
-                        materials.add(entries.getKey());
-                    }
-                }
-
-                Material material = materials.get(random.nextInt(materials.size())).parseMaterial();
+                Material[] materials = upgradesParsed.get(IridiumSkyblock.getInstance().getIslandManager().getIslandUpgrade(island.get(), "generator").getLevel())[newMaterial.equals(XMaterial.BASALT) ? 1 : 0];
```
dlsf, line 70: This line should probably be split up

src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
```diff
@@ -1,42 +1,78 @@
...
+                Material material = materials[random.nextInt(materials.length)];
+                if (material == Material.COBBLESTONE && event.getBlock().getType() == Material.STONE)
+                    material = Material.STONE;
```
dlsf, line 73: No more cobblestone? I don't play Skyblock myself, is that normal? 👀

dlsf (changes requested):

src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
PeachesMLG: Took me a while to realise what this was doing aswell, there should probabbly be a comment or something xD

What its doing is if there is a stone generator, and the new material is cobblestone, instead of setting it to cobblestone it should set it to stone instead since it was actually a stone generator normal cobblegens wont be affected

dlsf: Oh, alright, my bad

src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
Nubebuster: its shorter than line 67

src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
Nubebuster: True, true

src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
Nubebuster: Oops

PR #158, opened 2021-06-15
Cleanup & Optimise the ore generation, Add stone generators

dlsf:
My attempt at optimizing and cleaning up the generator. Alternative to #154


dlsf:
@Nubebuster I'd appreciate your opinion. I don't want to steal your idea but I just don't really like your implementation, it's pretty messy

Nubebuster:
Very clean!

PR #163, opened 2021-06-16
Fix main command not being executable in console

dlsf:


dlsf:
Should we execute the help command instead? I guess that would make sense

PR #165, opened 2021-06-16
Fix islandrewards not being saved on databases with a too small default column width

### LINKED ISSUE #164: IslandRewards exceed data storage limit in mysql
The table manager does not know how big data can be and island rewards exceed the expected data size. Right now, the max data size for island rewards is 255 characters. This causes the following error:
https://hastebin.com/acetocibol.sql

Nubebuster:
Raised column width default to 2048 to accomodate for more data. This data includes reward item lores and reward commands which can be added by a user through config. Therefore this limit can still be reached, but the user will have to change the limit in their database themself if they reach it.
closes #164

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/database/IslandReward.java
+++ b/src/main/java/com/iridium/iridiumskyblock/database/IslandReward.java
@@ -22,7 +22,7 @@ public final class IslandReward extends IslandData {
     @DatabaseField(columnName = "id", generatedId = true, canBeNull = false)
     private int id;
 
-    @DatabaseField(columnName = "reward", canBeNull = false)
+    @DatabaseField(columnName = "reward", canBeNull = false, width = 2048)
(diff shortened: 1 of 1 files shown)


dlsf (approved):
It would be good to know if OrmLite automatically adjusts the size of the column if the table has already been created. If it doesn't we should write some kind of converter.

dlsf:
Additionally, we should look if this same issue could also occur in other tables

PeachesMLG:
Is this compatable with older versions? ik if you add a new collum orm will complain saying it cant find that collum, does it complain about anything here?

Nubebuster:
> Is this compatable with older versions? ik if you add a new collum orm will complain saying it cant find that collum, does it complain about anything here?

This is compatible. People with an old database will have to manually change the width currently.

dlsf:
We should write something that can change the length automatically before merging this

PR #186, opened 2021-06-18
Fix warp messages & Fix ArrayIndexOutOfBoundsException

### LINKED ISSUE #178: When do /is deletewarp but i can't remake this error :/
https://pastebin.com/RrbX7EyZ

dlsf:
Fixes #178


PeachesMLG:
Id rather keep the syntax /is warp <name> but we should probabbly add a prefix to the other ones

dlsf:
I'm confused, what do you mean with that?

dlsf:
We should distinguish between optional and required arguments. Looks pretty weird sometimes (for instance for /is bank) right now

PeachesMLG:
> We should distinguish between optional and required arguments. Looks pretty weird sometimes (for instance for /is bank) right now

Ye i believe the convention is <> for required and () for optional ill add that tmr

PR #191, opened 2021-06-19
Player head multi thread

### LINKED ISSUE #185: /visit freezes the server
/visit does database operations on the main thread when populating the VisitGUI

Nubebuster:
Fixes #185 when you add a player head with a head owner into an inventory, bukkit performs a head data lookup on the main server thread. This blocks the main thread for up to two seconds per lookup. This patch adds SkinUtils where you can lookup the texture before adding the player head to an inventory. SkinUtils also caches previsouly looked up textures. 

Let me explain what the code does in VisitGUI as an example;
With my patch the code in VisitGUI the checks if the head data for the player owner exists in cache. If it does, the head is added to the inventory. If it does not exist, it puts a head with a configurable head data and adds the item to a list to later perform a skin lookup. For every skin lookup a new async thread is created. This is needed because looking up 20 heads without multithreading can take up to 30 seconds. Once the head data is fetched from mojang's api, the head with the player head data is added to the inventory using a sync task. (not a thread) This means for an entire GUI it takes about 2 seconds to load the player skins, and it does not block the main thread.

Users can still configure a custom head data in the Inventories config if they want.


Nubebuster:
This pull request also removes a typo that got in from one of my earlier  pull requests. It was an extra '+' symbol that didn't do any harm but also did not server a purpose

src/main/java/com/iridium/iridiumskyblock/gui/InvitesGUI.java
```diff
@@ -33,36 +38,56 @@
     public InvitesGUI(@NotNull Island island) {
         this.island = island;
         invites = new HashMap<>();
+        inventory = Bukkit.createInventory(this, IridiumSkyblock.getInstance().getInventories().islandInvitesGUI.size, StringUtils.color(IridiumSkyblock.getInstance().getInventories().islandInvitesGUI.title));
```
dlsf, line 41: Invites can be added. All other inventories update when their content changes

src/main/java/com/iridium/iridiumskyblock/gui/IslandTopGUI.java
```diff
@@ -45,24 +55,43 @@ public IslandTopGUI() {
...
+                    if (querySkin) Bukkit.getScheduler().runTaskAsynchronously(IridiumSkyblock.getInstance(), () -> {
+                        String headData = SkinUtils.queryHeadData(island.getOwner().getUuid());
+                        if (headData == null) return;
+                        initialItem.headData = headData;
+                        ItemStack itemStackWithSkin = ItemStackUtils.makeItem(initialItem, placeholderList);
+                        Bukkit.getScheduler().runTask(IridiumSkyblock.getInstance(), () -> inventory.setItem(slot, itemStackWithSkin));
+                    });
+                } else {
+                    inventory.setItem(slot, ItemStackUtils.makeItem(IridiumSkyblock.getInstance().getInventories().islandTopGUI.item, new PlaceholderBuilder().applyIslandPlaceholders(island).build()));
+                }
```
dlsf, line 94: Duplicated code, is there no way to unify this?

src/main/java/com/iridium/iridiumskyblock/utils/SkinUtils.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private static HashMap<UUID, String> cache = new HashMap<>();
+
+    public static String getHeadData(UUID uuid) {
+        return cache.get(uuid);
+    }
+
+    public static String queryHeadData(UUID uuid) {
+        try {
+            Gson g = new Gson();
```
dlsf, line 25: You know my opinion about such short names by now, I guess :D

dlsf (changes requested):
Much duplicated, some pretty long methods and inventories that no longer update in "real time".

I think there are some things we should discuss

src/main/java/com/iridium/iridiumskyblock/gui/IslandTopGUI.java
PeachesMLG: How about making a pr on [IridiumCore](https://github.com/Iridium-Development/IridiumCore) adding a method getPlayerHead(Player) or smthn? if not ill add that when I get home on wednesday

src/main/java/com/iridium/iridiumskyblock/utils/SkinUtils.java
Nubebuster: I'm not the author of that code

dlsf: But you can change it 🤔

PeachesMLG:
Would something like this be better? https://github.com/Iridium-Development/IridiumSkyblock/pull/197
I slightly changed the SkinUtils class and just made addconent call async

PR #214, opened 2021-06-25
Add cooldowns

dlsf:
Doesn't work yet because of Jackson issues (https://github.com/Iridium-Development/IridiumCore/pull/11). Also needs some testing

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
+++ b/src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
@@ -156,7 +156,7 @@ public class Messages {
     public String invalidBiome = "%prefix% &7That Biome doesn't exist.";
     public String changedBiome = "%prefix% &7You have changed your island biome to %biome%!";
     public String borderColorDisabled = "%prefix% &7That border color has been disabled.";
-    public String activeCooldown = "%prefix% &7You cannot do that due to a running cooldown, please wait %hours% hours, %minutes% minutes, %seconds% seconds!";
+    public String activeCooldown = "%prefix% &7You cannot do that due to a running cooldown, please wait %hours% hour(s), %minutes% minute(s), %seconds% second(s)!";
     public String yes = "Yes";
     public String no = "No";
     public String none = "None";
--- a/src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
+++ b/src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
@@ -154,6 +154,7 @@ public class Messages {
     public String islandChatEnabled = "%prefix% &7Island Chat Enabled!";
     public String islandChatDisabled = "%prefix% &7Island Chat Disabled!";
     public String borderColorDisabled = "%prefix% &7That border color has been disabled.";
+    public String activeCooldown = "%prefix% &7You cannot do that due to a running cooldown, please wait %duration%!";
     public String yes = "Yes";
     public String no = "No";
     public String none = "None";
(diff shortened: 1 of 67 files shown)


dlsf:
Fixes #87 now

src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
```diff
@@ -154,6 +154,7 @@
     public String islandChatEnabled = "%prefix% &7Island Chat Enabled!";
     public String islandChatDisabled = "%prefix% &7Island Chat Disabled!";
     public String borderColorDisabled = "%prefix% &7That border color has been disabled.";
+    public String activeCooldown = "%prefix% &7You cannot do that due to a running cooldown, please wait %duration%!";
```
PeachesMLG, line 157: feel like it would make more sence just to merge this and the duration format?

src/main/java/com/iridium/iridiumskyblock/managers/CooldownProvider.java
```diff
@@ -0,0 +1,92 @@
...
+        return cooldownTimes.get(t);
+    }
+
+    /**
+     * Reset the cooldown of the specified entity with the duration of this CooldownProvider.
+     * {@link CooldownProvider#isOnCooldown(Object)} will return true after this.
+     *
+     * @param t The entity which should be checked
+     */
+    public void resetCooldown(T t) {
```
PeachesMLG, line 62: not sure I like this naming, reset cooldown gives the impression that you are resetting the cooldown for the player and therefore they will no longer be on cooldown

PeachesMLG (changes requested):
Ill do some testing later today

src/main/java/com/iridium/iridiumskyblock/managers/CooldownProvider.java
dlsf: What would you suggest?

src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
dlsf: Possible but we also have the time format in the config.yml

PeachesMLG: Ye but the time in config.yml is for multiple stuff and I just feel like it makes more sence to include this into active cooldown

src/main/java/com/iridium/iridiumskyblock/managers/CooldownProvider.java
PeachesMLG: idk im terrible at naming stuff, just call it `public void i()`

PeachesMLG: apply cooldown? feel like that makes way more sence

dlsf: Sounds worse

PeachesMLG: How so?

dlsf: In my understanding cooldowns are not something that can be applied

PeachesMLG: addCooldown?

dlsf: That was about what I initially planned but the cooldown is not always added because it also resets

dlsf: I pick resetCooldown

dlsf: I'll think about it

PR #228, opened 2021-07-02
Fix entities teleporting to the wrong world

Gaider10:
Fixes entities trying to teleport to normal worlds when going through portals

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/listeners/EntityPortalListener.java
+++ b/src/main/java/com/iridium/iridiumskyblock/listeners/EntityPortalListener.java
@@ -1,0 +1,43 @@
+package com.iridium.iridiumskyblock.listeners;
+
+import com.iridium.iridiumskyblock.IridiumSkyblock;
+import com.iridium.iridiumskyblock.managers.IslandManager;
+import com.iridium.iridiumskyblock.utils.LocationUtils;
+import org.bukkit.World;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.entity.EntityPortalEvent;
+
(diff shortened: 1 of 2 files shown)


src/main/java/com/iridium/iridiumskyblock/listeners/EntityPortalListener.java
```diff
@@ -0,0 +1,43 @@
...
+public class EntityPortalListener implements Listener {
+
+    @EventHandler
+    public void onEntityPortal(EntityPortalEvent event) {
+        if(event.getTo() != null && event.getTo().getWorld() != null) {
+            IslandManager islandManager = IridiumSkyblock.getInstance().getIslandManager();
+            IridiumSkyblock.getInstance().getIslandManager().getIslandViaLocation(event.getFrom()).ifPresent(island -> {
+                World worldFrom = event.getFrom().getWorld();
+                String worldToName = event.getTo().getWorld().getName();
+                if ((worldToName.equals("world_nether") && Objects.equals(worldFrom, islandManager.getWorld())) || (worldToName.equals("world") && Objects.equals(worldFrom, islandManager.getNetherWorld()))) {
```
MertUnverdi, line 22: Every world name is "world"?
Make it via Bukkit#getWorld

Gaider10: I'm pretty sure that there can't be multiple worlds with the same name

dlsf: I guess he meant that users may not use these hardcoded world names

MertUnverdi:
stupid peaches.

PR #229, opened 2021-07-02
Add support for inventory title placeholders

dlsf:
They are already used in the warps GUI but not supported

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/gui/GUI.java
+++ b/src/main/java/com/iridium/iridiumskyblock/gui/GUI.java
@@ -2,24 +2,48 @@ package com.iridium.iridiumskyblock.gui;
 
 import com.iridium.iridiumcore.utils.StringUtils;
 import com.iridium.iridiumskyblock.IridiumSkyblock;
+import com.iridium.iridiumskyblock.PlaceholderBuilder;
 import com.iridium.iridiumskyblock.configs.inventories.NoItemGUI;
-import lombok.AllArgsConstructor;
+import com.iridium.iridiumskyblock.database.Island;
 import lombok.NoArgsConstructor;
 import org.bukkit.Bukkit;
 import org.bukkit.event.inventory.InventoryClickEvent;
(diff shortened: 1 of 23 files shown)


src/main/java/com/iridium/iridiumskyblock/gui/GUI.java
```diff
@@ -2,24 +2,48 @@
...
+
+    /**
+     * The default constructor.
+     *
+     * @param noItemGUI The NoItemGUI of this GUI
+     * @param island    The island of this GUI. Can be null
+     */
+    public GUI(@NotNull NoItemGUI noItemGUI, @Nullable Island island) {
+        if (island != null) {
+            this.noItemGUI = new NoItemGUI(
```
PeachesMLG, line 33: Not sure if this will work since the noItemGUI might not actually be a noItemGUI, everything just extends from a noItemGUI

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/gui/GUI.java
PeachesMLG: would it be better to just edit the getInventory method and do like List<Placeholder> placeholders = island==null?Collections.emptyList:new PlaceholderBuilder().applyIslandPlaceholders().build()

dlsf: Good point

src/main/java/com/iridium/iridiumskyblock/gui/GUI.java
```diff
@@ -20,7 +21,9 @@
 @NoArgsConstructor
 public abstract class GUI implements InventoryHolder {
 
-    private @Nullable NoItemGUI noItemGUI;
+    private NoItemGUI noItemGUI;
+    @JsonIgnore
```
PeachesMLG, line 25: I mean this isnt needed since this class doesnt get serialised

PR #231, opened 2021-07-02
Fix ForeignIslandTableManager NullPointerException

### LINKED ISSUE #222: Console error IridiumSkyblock-3.0.4 || 3.0.5
**Issue :**
When I use the /is menu it generates this in console

**Server version:** 1.16.5 purpurclip
**Plugin version:** 3.0.4 - 3.0.5 (The same thing happens to me with both versions)

Server log: https://pastebin.com/w7KEqn5Q

dlsf:
Fixes #222


PeachesMLG:
I dont see why they would be null but sure

PR #236, opened 2021-07-03
Made inventories sync

PeachesMLG:
Not sure if I like this or not, feel free to look at it and yell at me


PeachesMLG:
Its also kinda dum cuz we also get every texture for every island owner one by one which can take ages when you have alot of islands

dlsf (changes requested):
Looks good otherwise

PR #246, opened 2021-07-04
Placeholder support to island chat

MertUnverdi:
Add User Island chat toggle event (from #suggestions)

Add placeholders to island chat

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/api/UserChatToggleEvent.java
+++ b/src/main/java/com/iridium/iridiumskyblock/api/UserChatToggleEvent.java
@@ -1,0 +1,43 @@
+package com.iridium.iridiumskyblock.api;
+
+import com.iridium.iridiumskyblock.database.Island;
+import com.iridium.iridiumskyblock.database.User;
+import lombok.Getter;
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.NotNull;
+
+@Getter
+public class UserChatToggleEvent extends Event implements Cancellable {
+
+    private static final HandlerList handlers = new HandlerList();
+    private boolean cancelled;
+    private final Island island;
+    private final User user;
+
+    public UserChatToggleEvent(Island island, User user) {
+        this.island = island;
+        this.user = user;
+    }
+
+    @NotNull
+    public HandlerList getHandlers() {
+        return handlers;
+    }
+
+    @NotNull
+    public static HandlerList getHandlerList() {
+        return handlers;
+    }
+
+    @Override
+    public boolean isCancelled() {
+        return cancelled;
+    }
+
+    @Override
+    public void setCancelled(boolean b) {
+        cancelled = b;
+    }
+}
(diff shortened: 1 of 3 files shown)


src/main/java/com/iridium/iridiumskyblock/api/UserChatToggleEvent.java
```diff
@@ -0,0 +1,43 @@
...
+
+@Getter
+public class UserChatToggleEvent extends Event implements Cancellable {
+
+    private static final HandlerList handlers = new HandlerList();
+    private boolean cancelled;
+    private final Island island;
+    private final User user;
+
+    public UserChatToggleEvent(Island island, User user) {
```
PeachesMLG, line 19: This should also contain a state, like boolean chatEnabled, ye I know they can do user.getIslandChat or whatever but I feel like it should be exposed here aswell to make it easier

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/api/UserChatToggleEvent.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public boolean isCancelled() {
+        return cancelled;
+    }
+
+    @Override
+    public void setCancelled(boolean b) {
+        cancelled = b;
```
dlsf, line 41: Why b? D:
```suggestion
    public void setCancelled(boolean cancel) {
        cancelled = cancel;
```

src/main/java/com/iridium/iridiumskyblock/listeners/PlayerChatListener.java
```diff
@@ -16,30 +17,30 @@
 
 public class PlayerChatListener implements Listener {
 
-    @EventHandler
+    @EventHandler(ignoreCancelled = true)
```
dlsf, line 20: Ignoring cancelled event calls results in players being able to bypass mutes etc

dlsf (changes requested):

src/main/java/com/iridium/iridiumskyblock/listeners/PlayerChatListener.java
dlsf:
```suggestion
    @EventHandler
```

src/main/java/com/iridium/iridiumskyblock/api/UserChatToggleEvent.java
PeachesMLG: I assumed just to annoy you

src/main/java/com/iridium/iridiumskyblock/listeners/PlayerChatListener.java
PeachesMLG: I dont see how this would cause any issues? it just wont run our code if the event is cancelled, it basically just means that the players wont be able to use /is chat if they are muted

dlsf: Oh, I see, my bad

src/main/java/com/iridium/iridiumskyblock/api/UserChatToggleEvent.java
```diff
@@ -12,10 +12,12 @@
 
     private static final HandlerList handlers = new HandlerList();
     private boolean cancelled;
+    private final boolean isIslandChat;
     private final User user;
 
     public UserChatToggleEvent(User user) {
         this.user = user;
+        this.isIslandChat = user.isIslandChat();
```
PeachesMLG, line 20: You need to parse the new state into the event, this is the old state since this event gets called before the state changes

PR #258, opened 2021-07-09
Fix world issues

dlsf:
Fixes issues with missing world checks, e.g. for missions.
Also fixes border issues on respawn.

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
+++ b/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
@@ -8,6 +8,7 @@ import com.iridium.iridiumskyblock.database.IslandBooster;
 import com.iridium.iridiumskyblock.database.IslandUpgrade;
 import com.iridium.iridiumskyblock.database.User;
 import com.iridium.iridiumskyblock.managers.IslandManager;
+import java.util.Objects;
 import org.bukkit.Location;
 import org.bukkit.OfflinePlayer;
 import org.bukkit.World;
@@ -271,7 +272,7 @@ public class IridiumSkyblockAPI {
      * @since 3.0.7
      */
     public boolean isIslandWorld(World world) {
-        return getWorld() == world || getNetherWorld() == world || getEndWorld() == world;
+        return Objects.equals(getWorld(), world) || Objects.equals(getNetherWorld(), world) || Objects.equals(getEndWorld(), world);
     }
 
 }
--- a/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
+++ b/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
@@ -263,4 +263,15 @@ public class IridiumSkyblockAPI {
         return iridiumSkyblock.getIslandManager().getEndWorld();
     }
 
+    /**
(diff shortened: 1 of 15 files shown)


src/main/java/com/iridium/iridiumskyblock/listeners/PlayerMoveListener.java
```diff
@@ -17,6 +18,10 @@
 
     @EventHandler
     public void onPlayerMove(PlayerMoveEvent event) {
+        if (!IridiumSkyblockAPI.getInstance().isIslandWorld(event.getFrom().getWorld()) && event.getTo() != null && !IridiumSkyblockAPI.getInstance().isIslandWorld(event.getTo().getWorld())) {
```
PeachesMLG, line 21: This is bad, even tho they arnt in the skyblock world, still means they should have the teleport dont move thingy also IslandManager#getIslandViaLocation includes the world checks

src/main/java/com/iridium/iridiumskyblock/listeners/PlayerRespawnListener.java
```diff
@@ -17,8 +17,8 @@ public void onRespawn(PlayerRespawnEvent event) {
             user.getIsland().ifPresent(island -> {
                 if (!island.isInIsland(event.getRespawnLocation())) {
                     event.setRespawnLocation(LocationUtils.getSafeLocation(island.getHome(), island));
+                    PlayerUtils.sendBorder(event.getPlayer(), island);
```
PeachesMLG, line 20: this is wrong? if should send the border regardless since either way the respawn location will end up being in the island?

src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
```diff
@@ -263,4 +263,15 @@ public World getEndWorld() {
...
 
+    /**
+     * Returns whether or not the specified world is from IridiumSkyblock.
+     *
+     * @param world Thw world that should be checked
+     * @return true if it is a world used by IridiumSkyblock
+     * @since 3.0.7
+     */
+    public boolean isIslandWorld(World world) {
+        return getWorld() == world || getNetherWorld() == world || getEndWorld() == world;
```
PeachesMLG, line 274: I think we should have this method in IslandManager and link this method to the IslandManager's method since thats what all the other stuff does

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
dlsf: It's more of a utility method for checking if IridiumSkyblockAPI#getWorld(), IridiumSkyblockAPI#getNetherWorld() or IridiumSkyblockAPI#getEndWorld() equal to another world. It makes more sense to be here in my opinion

src/main/java/com/iridium/iridiumskyblock/listeners/PlayerRespawnListener.java
dlsf: You are right, but the way it is now, the border is always sent even if the player doesn't respawn on his island / in the island worlds

dlsf: I tried to fix it, please review 👀

PR #268, opened 2021-07-12
Fix unsupported biome selection

dlsf:


PeachesMLG (changes requested):
Looks good, you also want to do this on the BiomeCommand auto complete stuff

PR #272, opened 2021-07-12
Add Plugin config load event

MertUnverdi:


src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockReloadEvent.java
```diff
@@ -0,0 +1,20 @@
+package com.iridium.iridiumskyblock.api;
+
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.NotNull;
+
+public class IridiumSkyblockReloadEvent extends Event {
+    private static final HandlerList handlers = new HandlerList();
```
dlsf, line 8: If you're putting the empty line at the end of the class...
```suggestion
public class IridiumSkyblockReloadEvent extends Event {

    private static final HandlerList handlers = new HandlerList();
```

MertUnverdi: No i didn't put that line -_-

dlsf (approved):
That was not what I meant but whatever

PR #279, opened 2021-07-13
ignore registering listener while oreupgrade is disabled

MertUnverdi:
why we're running this event while upgrade is disabled?
BlockFormListener#41 runs getIsland method we don't need it, this will more effiencient


dlsf:
I'm not sure if this is needed... I mean... We cache the islands, have a fast way to get islands by locations now and most people don't actually disable the generator - so it's pretty much just more complicated code.

PeachesMLG:
I dont like this for a few reasons:

1) If they then enable it on /is reload it wont work
2) Somone might want to disable the upgrade but still have an ore generator
3) I dont have a 3rd reason but things look better in 3's

PR #283, opened 2021-07-16
Fix missing booster & upgrade purchase message

dlsf:


dlsf:
https://github.com/Iridium-Development/IridiumColorAPI/pull/9 should be merged first

src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
```diff
@@ -147,6 +147,8 @@
     public String noSuchItem = "%prefix% &7You don't have this item!";
     public String successfullyBought = "%prefix% &7You have successfully bought %amount%x %item% &r&7for $%vault_cost% and %crystal_cost% Crystals.";
     public String successfullySold = "%prefix% &7You have successfully sold %amount%x %item% &r&7for $%vault_reward% and %crystal_reward% Crystals.";
+    public String successfullyBoughtBooster = "%prefix% &7You have successfully bought the %booster% Booster &r&7for $%vault_cost% and %crystal_cost% Crystals.";
```
PeachesMLG, line 150: The &7&r isnt needed here? since its already &7?

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/commands/BoostersCommand.java
```diff
@@ -54,6 +54,16 @@ public boolean execute(CommandSender sender, String[] args) {
                             islandBooster.setTime(LocalDateTime.now().plusSeconds(booster.time));
                             IslandLog islandLog = new IslandLog(island.get(), LogAction.BOOSTER_PURCHASE, user, null, 0, boosterName);
                             IridiumSkyblock.getInstance().getDatabaseManager().getIslandLogTableManager().addEntry(islandLog);
+
+                            player.sendMessage(
+                                StringUtils.color(
+                                    IridiumSkyblock.getInstance().getMessages().successfullyBoughtBooster
+                                        .replace("%prefix%", IridiumSkyblock.getInstance().getConfiguration().prefix)
+                                        .replace("%booster%", IridiumSkyblock.getInstance().getBoosterList().get(islandBooster.getBooster()).item.displayName)
```
PeachesMLG, line 62: Just set this to entry.key with some Capitalisation magic?

src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
dlsf: It is because the booster name can be bold for example. You could argue whether or not the &r should be in front of `Booster` though.

src/main/java/com/iridium/iridiumskyblock/commands/BoostersCommand.java
dlsf: We talked about this on Discord, I guess? Are you fine with this solution?

src/main/java/com/iridium/iridiumskyblock/configs/Messages.java
dlsf: But I have to say that it looked pretty good in chat

MertUnverdi:
why is still not merged??

PR #285, opened 2021-07-18
Add Obsidian -> Lava Bucket

Oribuin:
This is a useful feature for those absolute degenerates that dont know how to make a cobble gen, who doesnt know how to make a cobblegen in 2021? anyway lets not bully those bakas anymore.

It can be used to covert obsidian back into lava whenever they do mess up some how. I wrote this for 1.17 so it likely isnt that backwards compatible but whatever, Suck it old verisons

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/listeners/BucketListener.java
+++ b/src/main/java/com/iridium/iridiumskyblock/listeners/BucketListener.java
@@ -5,12 +5,19 @@ import com.iridium.iridiumskyblock.IridiumSkyblock;
 import com.iridium.iridiumskyblock.PermissionType;
 import com.iridium.iridiumskyblock.database.Island;
 import com.iridium.iridiumskyblock.database.User;
+import com.iridium.iridiumskyblock.managers.IslandManager;
+import org.bukkit.Bukkit;
+import org.bukkit.Material;
+import org.bukkit.block.Block;
 import org.bukkit.entity.Player;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import org.bukkit.event.player.PlayerBucketEmptyEvent;
 import org.bukkit.event.player.PlayerBucketEvent;
 import org.bukkit.event.player.PlayerBucketFillEvent;
+import org.bukkit.event.player.PlayerInteractEvent;
+import org.bukkit.inventory.EquipmentSlot;
+import org.bukkit.inventory.ItemStack;
 
 import java.util.Optional;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/iridium/iridiumskyblock/listeners/BucketListener.java
```diff
@@ -40,4 +48,44 @@ public void onBucketEvent(PlayerBucketEvent event) {
...
+        if (!islandOptional.isPresent())
+            return;
+
+        final User user = IridiumSkyblock.getInstance().getUserManager().getUser(event.getPlayer());
+
+        if (!manager.getIslandPermission(islandOptional.get(), user, PermissionType.BUCKET))
+            return;
+
+
+        event.getClickedBlock().setType(Material.AIR);
```
MertUnverdi, line 84: it will work to all obsidians, u should check this obsidian from an lava

MertUnverdi: so we don't player player break a nether portal with a bucket

PeachesMLG (changes requested):
Needs a config option to disable/enable this

src/main/java/com/iridium/iridiumskyblock/listeners/BucketListener.java
PeachesMLG: Perhaps make a database class that saves all the failed ones to a database :thinking:

Oribuin: From what I know there's no spigot method to check if a block was placed or formed

PeachesMLG: There is, Block form listener check if its new state is obsidian or not

PeachesMLG: Then add to a database or hashmap or something

Oribuin: Yeah but that's in the BlockFormListener event. There is no way to check a block's potential previous state outside of the event. The only alternative is to cache all the current formed blocks in the plugin. 

Saving all the formed blocks seems incredibly unnecessary and just caching it without saving it would reset every restart

PeachesMLG: Its not going to save all formed blocks, only the ones that are obsidian, and putting it in a hashmap would reset it every restart but people would usually click it with a bucket within like 10 seconds to get it back so :shrug:

PR #296, opened 2021-07-19
Make confirmation gui yes-no item slots changeable

MertUnverdi:


dlsf (approved):

PeachesMLG:
Would this work with preexisting configs? the slots are null by default are they not?

PR #300, opened 2021-07-21
Fix block limit upgrade

xReveres:
Fixes enabling/disabling block limit upgrade

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/listeners/BlockPlaceListener.java
+++ b/src/main/java/com/iridium/iridiumskyblock/listeners/BlockPlaceListener.java
@@ -45,7 +45,7 @@ public class BlockPlaceListener implements Listener {
 
         IslandBlocks islandBlocks = IridiumSkyblock.getInstance().getIslandManager().getIslandBlock(island.get(), material);
         
-        if(IridiumSkyblock.getInstance().getUpgrades().blockLimitUpgrade.enabled){
+        if(IridiumSkyblock.getInstance().getUpgrades().blockLimitUpgrade.enabled) {
             int limitUpgradeLevel = IridiumSkyblock.getInstance().getIslandManager().getIslandUpgrade(island.get(), "blocklimit").getLevel();
             int blockLimit = IridiumSkyblock.getInstance().getUpgrades().blockLimitUpgrade.upgrades.get(limitUpgradeLevel).limits.getOrDefault(material, 0);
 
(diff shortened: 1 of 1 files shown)


PeachesMLG:
Not sure I like this, mabey they dont want a block limit upgrade but still want block limits? in which case they can change the limits for the level 1 upgrade, if they dont want limits or an upgrade then they just need to remove all the limits from level 1?

xReveres:
I thought property "enabled" is for whole upgrade not for only display icon in upgrade GUI, but also you have right, so maybe we can add new property "displayInGUI: true/false" or add support for "slot: -1" if they dont want display it, but want limit/block something?

src/main/java/com/iridium/iridiumskyblock/listeners/BlockPlaceListener.java
```diff
@@ -44,15 +44,19 @@ public void onBlockPlaceEvent(BlockPlaceEvent event) {
         }
 
         IslandBlocks islandBlocks = IridiumSkyblock.getInstance().getIslandManager().getIslandBlock(island.get(), material);
-        int limitUpgradeLevel = IridiumSkyblock.getInstance().getIslandManager().getIslandUpgrade(island.get(), "blocklimit").getLevel();
-        int blockLimit = IridiumSkyblock.getInstance().getUpgrades().blockLimitUpgrade.upgrades.get(limitUpgradeLevel).limits.getOrDefault(material, 0);
+        
+        if(IridiumSkyblock.getInstance().getUpgrades().blockLimitUpgrade.enabled) {
```
dlsf, line 48:
```suggestion
        if (IridiumSkyblock.getInstance().getUpgrades().blockLimitUpgrade.enabled) {
```

dlsf (changes requested):
Very important

dlsf (approved):

PeachesMLG:
> I thought property "enabled" is for whole upgrade not for only display icon in upgrade GUI, but also you have right, so maybe we can add new property "displayInGUI: true/false" or add support for "slot: -1" if they dont want display it, but want limit/block something?

I dont see the issue with the system we have no, its the same for every upgrade. For example the size upgrade, if enabled is true its saying we can upgrade it, if its false we cant upgrade it, but we will still have a size, its just you cant upgrade the size so it uses level 1 as a default, same with every other upgrade?

xReveres:
Yes you have right, my mistake.
My idea was to make a little "performance" to omit few lines of code and checks.

PR #307, opened 2021-07-22
Fix database XMaterial loading

### LINKED ISSUE #298: Cannot get enum value of 'GRASS_PATH'
**Describe the bug**
plugin no worky with db )

**To Reproduce**
1. Use any version above 3.0.3
2. Load the db file attached
3. See error

**Expected behavior**
for plugin to work

Versions 3.0.3+, Minecraft 1.16.5

**Screenshots**

**Logs**
If there are error messages in your console, please send them via [This](https://paste.gg/ )
https://paste.gg/p/anonymous/09568d2eae4e43129c7e457a6ee19d83

**Additional context**
Resetting the db file fixes this

xReveres:
Fixes #298 and other future XMaterial changes

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/database/types/XMaterialType.java
+++ b/src/main/java/com/iridium/iridiumskyblock/database/types/XMaterialType.java
@@ -7,30 +7,28 @@ import com.j256.ormlite.field.types.StringType;
 import java.sql.SQLException;
 import java.util.Optional;
 
-public class XMaterialType extends StringType{
+public class XMaterialType extends StringType {
 
-    private static final XMaterialType singleTon = new XMaterialType();
+    private static final XMaterialType instance = new XMaterialType();
 
     public static XMaterialType getSingleton() {
-		return singleTon;
-	}
+        return instance;
+    }
 
     protected XMaterialType() {
         super(SqlType.STRING, new Class<?>[] { XMaterial.class });
     }
 
     @Override
-	public Object sqlArgToJava(FieldType fieldType, Object sqlArg, int columnPos) throws SQLException {
+    public Object sqlArgToJava(FieldType fieldType, Object sqlArg, int columnPos) throws SQLException {
         String value = (String) super.sqlArgToJava(fieldType, sqlArg, columnPos);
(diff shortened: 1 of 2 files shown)


src/main/java/com/iridium/iridiumskyblock/database/types/XMaterialType.java
```diff
@@ -0,0 +1,37 @@
...
+        return null;
+    }
+
+    @Override
+	public Object javaToSqlArg(FieldType fieldType, Object obj) throws SQLException {
+        XMaterial material = (XMaterial) obj;
+        return super.javaToSqlArg(fieldType, material.toString());
+    }
+    
+}
```
dlsf, line 37: The indentation in this class is something else
```suggestion
public class XMaterialType extends StringType {

    private static final XMaterialType instance = new XMaterialType();

    public static XMaterialType getSingleton() {
        return instance;
    }

    protected XMaterialType() {
        super(SqlType.STRING, new Class<?>[] { XMaterial.class });
    }

    @Override
    public Object sqlArgToJava(FieldType fieldType, Object sqlArg, int columnPos) throws SQLException {
        String value = (String) super.sqlArgToJava(fieldType, sqlArg, columnPos);
        Optional<XMaterial> material = XMaterial.matchXMaterial(value);
        return material.orElse(null);
    }

    @Override
    public Object javaToSqlArg(FieldType fieldType, Object object) throws SQLException {
        XMaterial material = (XMaterial) object;
        return super.javaToSqlArg(fieldType, material.toString());
    }
    
}

```

dlsf (changes requested):
Also, I don't quite get the point of that singleton, is it required by OrmLite?

xReveres:
I inspired from this
https://github.com/j256/ormlite-core/blob/master/src/main/java/com/j256/ormlite/field/types/SerializableType.java

dlsf:
I see. Did you test if these changes actually work? OrmLite sometimes just ignores stuff that has been configured

xReveres:
Yes I tested it 2 times, for example I manually changed in database from "DIRT_PATH" to "GRASS_PATH" and works fine

dlsf (approved):

PR #314, opened 2021-07-22
Disable island fly on other worlds

### LINKED ISSUE #312: Island Flight works everywhere
**Describe the bug**
Players keep island fly in every world if they use /island fly.

**To Reproduce**
1. Use '/island fly' on island.
2. Use any teleportation command (/spawn, /warp, /tpa)
3. Boom, you have fly somewhere you shouldn't.

**Expected behavior**
Flight is toggled off with any teleportation command.

1.17.1, plugin 3.0.7

MertUnverdi:
fixes #312

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/database/User.java
+++ b/src/main/java/com/iridium/iridiumskyblock/database/User.java
@@ -7,7 +7,9 @@ import com.j256.ormlite.table.DatabaseTable;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
+import org.bukkit.Bukkit;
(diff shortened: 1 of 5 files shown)


dlsf:
I think that the main issue is that flight is not disabled when switching worlds and that the timer continues to run out

MertUnverdi:
> I think that the main issue is that flight is not disabled when switching worlds and that the timer continues to run out

may other users still on island also we can't stop timer :(

dlsf:
But we can add condition to its logic

dlsf:
I guess we could also just disable flight every time someone switches to a non IridiumSkyblock world + block /is fly in these

PR #327, opened 2021-07-29
Per-Placeholder Default Values

Oribuin:
This adds the ability for people to customize the default values for each individual placeholders instead of a global one.

This will use placeholders.yml

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/configs/Placeholders.java
+++ b/src/main/java/com/iridium/iridiumskyblock/configs/Placeholders.java
@@ -6,44 +6,45 @@ import com.iridium.iridiumcore.dependencies.fasterxml.annotation.JsonIgnorePrope
 public class Placeholders {
 
     // Island Placeholders.
-    public String island_name = "N/A";
-    public String island_owner = "N/A";
-    public String island_rank = "0";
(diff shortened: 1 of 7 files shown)


PeachesMLG (approved):
Looks good to me, but im dum so @dlsf ?

src/main/java/com/iridium/iridiumskyblock/configs/Placeholders.java
```diff
@@ -0,0 +1,49 @@
+package com.iridium.iridiumskyblock.configs;
+
+import com.iridium.iridiumcore.dependencies.fasterxml.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Placeholders {
+
+    // Island Placeholders.
+    public String island_name = "N/A";
```
dlsf, line 9: I'm afraid this isn't python, please use lowerCamelCase 👀

dlsf (changes requested):

MertUnverdi (approved):
we don't need new config files but peaches is dum

dlsf (approved):
No, Ry is dum

PeachesMLG:
Ry very dum

PR #330, opened 2021-07-30
add option to disable patreon message

Oribuin:
please add this, it took me a whole 30 seconds to create this feature and id really appreciate it if you didnt make me waste a whole 30s on making this change

also it can be kinda annoying so to save the whining from people in #support, just merge it, we'd all be better off

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerJoinQuitListener.java
+++ b/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerJoinQuitListener.java
@@ -26,7 +26,7 @@ public class PlayerJoinQuitListener implements Listener {
                 PlayerUtils.sendBorder(player, island)
         );
 
-        if (player.isOp()) {
+        if (player.isOp() && IridiumSkyblock.getInstance().getConfiguration().patreonMessage) {
             Bukkit.getScheduler().runTaskLater(IridiumSkyblock.getInstance(), () ->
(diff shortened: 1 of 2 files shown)


MertUnverdi:
close it dum, we won't remove that message (for some users should stay there)👍🏻

src/main/java/com/iridium/iridiumskyblock/listeners/PlayerJoinQuitListener.java
```diff
@@ -26,7 +26,7 @@ public void onJoin(PlayerJoinEvent event) {
                 PlayerUtils.sendBorder(player, island)
         );
 
-        if (player.isOp()) {
+        if (player.isOp() && IridiumSkyblock.getInstance().getConfiguration().patreonMessage) {
```
dlsf, line 29: This code can be simplified
```suggestion
        if (player.isOp() && true) {
```

Oribuin:
> close it dum, we won't remove that message (for some users should stay there)👍🏻

i will find your family and bonk you upside the noggin bruv

Oribuin:
the build failed because Esophose is moving the rosewood repo, I'm not a bad developer I promise

dlsf (changes requested):

src/main/java/com/iridium/iridiumskyblock/listeners/PlayerJoinQuitListener.java
MertUnverdi: LOOOL

PeachesMLG:
You know I really liked having this because it would annoy everyone, but fine >:C

PR #331, opened 2021-07-31
Added UserJoinEvent and missing `@Getter` in some other events

SoKnight:
I add missing and useful `UserJoinEvent` in the public API and made that calling in the `JoinCommand`.
Also the `UserLeaveEvent` has not a `@Getter` annotation above the class declaration, it also was fixed.
(sorry for any mistakes, it's my first PR here)

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/api/UserJoinEvent.java
+++ b/src/main/java/com/iridium/iridiumskyblock/api/UserJoinEvent.java
@@ -9,8 +9,6 @@ import org.bukkit.event.HandlerList;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.Optional;
-
 @Getter
 public class UserJoinEvent extends Event implements Cancellable {
     private static final HandlerList handlers = new HandlerList();
@@ -23,6 +21,10 @@ public class UserJoinEvent extends Event implements Cancellable {
         this.island = island;
         this.user = user;
         this.inviter = inviter;
+    }
+
+    public boolean isJoinedByInvite() {
+        return inviter != null;
     }
 
     @NotNull
--- a/src/main/java/com/iridium/iridiumskyblock/api/UserJoinEvent.java
+++ b/src/main/java/com/iridium/iridiumskyblock/api/UserJoinEvent.java
@@ -25,11 +25,6 @@ public class UserJoinEvent extends Event implements Cancellable {
         this.inviter = inviter;
     }
 
-    // make that Optional because the `inviter` can be null
-    public Optional<User> getInviter() {
-        return Optional.ofNullable(inviter);
-    }
-
     @NotNull
     public HandlerList getHandlers() {
         return handlers;
--- a/src/main/java/com/iridium/iridiumskyblock/api/UserJoinEvent.java
+++ b/src/main/java/com/iridium/iridiumskyblock/api/UserJoinEvent.java
@@ -7,17 +7,27 @@ import org.bukkit.event.Cancellable;
 import org.bukkit.event.Event;
 import org.bukkit.event.HandlerList;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Optional;
 
 @Getter
-public class UserLeaveEvent extends Event implements Cancellable {
+public class UserJoinEvent extends Event implements Cancellable {
     private static final HandlerList handlers = new HandlerList();
     private boolean cancelled;
     private final Island island;
     private final User user;
+    private final User inviter;
 
-    public UserLeaveEvent(Island island, User user) {
+    public UserJoinEvent(Island island, User user, @Nullable User inviter) {
         this.island = island;
         this.user = user;
+        this.inviter = inviter;
+    }
+
+    // make that Optional because the `inviter` can be null
+    public Optional<User> getInviter() {
+        return Optional.ofNullable(inviter);
     }
 
     @NotNull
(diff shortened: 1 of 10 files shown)


src/main/java/com/iridium/iridiumskyblock/commands/JoinCommand.java
```diff
@@ -64,6 +65,12 @@ public boolean execute(CommandSender sender, String[] args) {
                         return false;
                     }
 
+                    // PR: added UserJoinEvent
+                    Optional<User> inviter = islandInvite.map(IslandInvite::getInviter);
+                    UserJoinEvent userJoinEvent = new UserJoinEvent(island.get(), user, inviter.orElse(null));
```
MertUnverdi, line 70: Why the inviter is not offlinePlayerUser ? 
Also inviter how can be null?  Our invite command always has a inviter

MertUnverdi: Also can you change the offlinePlayer and offlinePlayerUser to targetPlayer and targetUser

SoKnight: > Also inviter how can be null? Our IslandInvite always has a inviter you can check on InviteCommand
This command also can be executed when someone staff joins to island, or not? I mean that he can do it without any invitation.

SoKnight: > Why the inviter is not offlinePlayerUser ?

Because `offlinePlayerUser` suggested by you don't describe what is that variable

src/main/java/com/iridium/iridiumskyblock/api/UserJoinEvent.java
```diff
@@ -0,0 +1,47 @@
...
+import com.iridium.iridiumskyblock.database.User;
+import lombok.Getter;
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Optional;
+
```
MertUnverdi, line 13: Remove this :panda:

src/main/java/com/iridium/iridiumskyblock/commands/JoinCommand.java
MertUnverdi: Um, well you right, may we can add a boolean like isForceJoin to joincommand and join event
What your thinks about this, good or weird ?

SoKnight: > Um, well you right, may we can add a boolean like isForceJoin to joincommand and join event
What your thinks about this, good or weird ?

Yes, we can add that as just a `inviter == null` check-method `isForceJoin` in the event class.
It's a current event logic i think.

src/main/java/com/iridium/iridiumskyblock/commands/JoinCommand.java
```diff
@@ -64,6 +65,12 @@ public boolean execute(CommandSender sender, String[] args) {
                         return false;
                     }
 
+                    // PR: added UserJoinEvent
```
dlsf, line 68: I don't really get the point of this comment and we haven't done that before. This would just become messy really quick. 
```suggestion
```

src/main/java/com/iridium/iridiumskyblock/commands/JoinCommand.java
```diff
@@ -64,6 +65,12 @@ public boolean execute(CommandSender sender, String[] args) {
...
+                    Optional<User> inviter = islandInvite.map(IslandInvite::getInviter);
+                    UserJoinEvent userJoinEvent = new UserJoinEvent(island.get(), user, inviter.orElse(null));
+                    Bukkit.getPluginManager().callEvent(userJoinEvent);
+                    if(userJoinEvent.isCancelled()) return false;
```
dlsf, line 72: Very important

```suggestion
                    if (userJoinEvent.isCancelled()) return false;
```

dlsf (changes requested):
Otherwise it looks fine to me

dlsf (approved):

src/main/java/com/iridium/iridiumskyblock/commands/JoinCommand.java
```diff
@@ -64,6 +65,11 @@ public boolean execute(CommandSender sender, String[] args) {
                         return false;
                     }
 
+                    Optional<User> inviter = islandInvite.map(IslandInvite::getInviter);
+                    UserJoinEvent userJoinEvent = new UserJoinEvent(island.get(), user, inviter.orElse(null));
+                    Bukkit.getPluginManager().callEvent(userJoinEvent);
+                    if (userJoinEvent.isCancelled()) return false;
```
MertUnverdi, line 71: @dlsf is should be true?
So, command is executed success but just ignored

SoKnight: > is should return true?

Oh, it's true, sorry, one minute :)

dlsf: No, technically it was not successful. I'd suggest to return false

MertUnverdi (approved):

src/main/java/com/iridium/iridiumskyblock/api/UserJoinEvent.java
```diff
@@ -0,0 +1,49 @@
...
+        return handlers;
+    }
+
+    @Override
+    public boolean isCancelled() {
+        return cancelled;
+    }
+
+    @Override
+    public void setCancelled(boolean b) {
```
PeachesMLG, line 46: Needs a better variable name

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/api/UserJoinEvent.java
SoKnight: Just copied from other event, but ok

PR #336, opened 2021-08-01
Delete duplicates from Database

PeachesMLG:
Not sure if I like this or not since the code is 99% the same for each method, feels like I can do this with a loop or something but im too dum to figure it out


dlsf:
There are also duplicate islands iirc

PeachesMLG:
I havent seen duplicate islands before? and that wouldnt really matter too much anyways since the id is a primary key so you can have duplicate islands but the id's have to be different

dlsf (approved):
Looks better

EvilInsid:
im testing iridium skyblock for my server, so far money randomly resets.... whats a good software to edit database files? and how often do you have to fix this?

PR #339, opened 2021-08-02
Island Biome Placeholders

Oribuin:
have I tested this? no, Will I? no

i wrote it so it has to work first time /s :)


PeachesMLG:
Should add biomes for nether and end aswell

PR #341, opened 2021-08-02
Anti-Island Member Kick

Oribuin:
The plugin will no longer kick island members and send them back to spawn when the island is set to private

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/commands/PrivateCommand.java
+++ b/src/main/java/com/iridium/iridiumskyblock/commands/PrivateCommand.java
@@ -44,9 +44,11 @@ public class PrivateCommand extends Command {
             island.get().setVisitable(false);
             int visitorCount = 0;
             List<IslandTrusted> islandTrusted = IridiumSkyblock.getInstance().getDatabaseManager().getIslandTrustedTableManager().getEntries(island.get());
+            final List<User> islandMembers = island.get().getMembers();
+
             for (User visitor : IridiumSkyblock.getInstance().getIslandManager().getPlayersOnIsland(island.get())) {
 
-                if (island.get().getMembers().contains(visitor))
+                if (islandMembers.contains(visitor))
                     continue;
 
                 if (visitor.getIsland().map(Island::getId).orElse(0) == island.get().getId() || islandTrusted.stream().anyMatch(trustedIsland -> trustedIsland.getUser().getUuid() == visitor.getUuid())) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/iridium/iridiumskyblock/commands/PrivateCommand.java
```diff
@@ -45,6 +45,10 @@ public boolean execute(CommandSender sender, String[] args) {
             int visitorCount = 0;
             List<IslandTrusted> islandTrusted = IridiumSkyblock.getInstance().getDatabaseManager().getIslandTrustedTableManager().getEntries(island.get());
             for (User visitor : IridiumSkyblock.getInstance().getIslandManager().getPlayersOnIsland(island.get())) {
+
+                if (island.get().getMembers().contains(visitor))
```
PeachesMLG, line 49: Island#getMembers should be saved as a variable somewhere to be more efficient, it wont make a huge difference but like ye

MertUnverdi: This already checks below line?

PeachesMLG: no island.getMembers performs a binary search on the users entries to get the members, so we should be saving something like Lisit<User> islandMembers = island.get().getMemebers() to make it more efficient

src/main/java/com/iridium/iridiumskyblock/commands/PrivateCommand.java
```diff
@@ -44,7 +44,13 @@ public boolean execute(CommandSender sender, String[] args) {
...
             int visitorCount = 0;
             List<IslandTrusted> islandTrusted = IridiumSkyblock.getInstance().getDatabaseManager().getIslandTrustedTableManager().getEntries(island.get());
+            final List<User> islandMembers = island.get().getMembers();
+
             for (User visitor : IridiumSkyblock.getInstance().getIslandManager().getPlayersOnIsland(island.get())) {
+
+                if (islandMembers.contains(visitor))
+                    continue;
+
                 if (visitor.getIsland().map(Island::getId).orElse(0) == island.get().getId() || islandTrusted.stream().anyMatch(trustedIsland -> trustedIsland.getUser().getUuid() == visitor.getUuid())) {
```
PeachesMLG, line 54: Slash just pointed out to me, we already check this here :eyes:

PR #351, opened 2021-08-05
Add cooldown for interaction messages

### LINKED ISSUE #349: Message spammed
**Describe the bug**
If you stand on a preasure plate on an island that you haven't permissions to use redstone, a message got spammed.

**To Reproduce**
1. Stand on a preasure plate on an island that your haven't permissions to use redstone

**Expected behavior**
Should display only one message

1.8.8 | 3.0.8

**Screenshots**
https://i.imgur.com/jOoKtuu.mp4

**Additional context**
I know that 1.8.8 are unsupported, but this issue is on all versions of mc

dlsf:
Fixes #349

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerInteractListener.java
+++ b/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerInteractListener.java
@@ -7,6 +7,8 @@ import com.iridium.iridiumskyblock.PermissionType;
 import com.iridium.iridiumskyblock.api.IridiumSkyblockAPI;
 import com.iridium.iridiumskyblock.database.Island;
 import com.iridium.iridiumskyblock.database.User;
+import com.iridium.iridiumskyblock.managers.CooldownProvider;
+import java.time.Duration;
 import org.bukkit.entity.Player;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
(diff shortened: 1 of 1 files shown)


src/main/java/com/iridium/iridiumskyblock/listeners/PlayerInteractListener.java
```diff
@@ -97,4 +112,13 @@ public void onPlayerInteractEntity(PlayerInteractEntityEvent event) {
         player.sendMessage(StringUtils.color(IridiumSkyblock.getInstance().getMessages().cannotInteractEntities.replace("%prefix%", IridiumSkyblock.getInstance().getConfiguration().prefix)));
     }
 
+    private boolean hasNoCooldown(Player player) {
+        if (cooldownProvider.isOnCooldown(player)) {
```
PeachesMLG, line 116: I think we should set the cooldown to like 0.1 seconds then apply the cooldown even if they are already on the cooldown.

So then it will only send the message once until they step of the pressure plate

lmk if this doesnt make sence cuz I dont understand what im saying half the time

dlsf: I thought that they should still know that they are doing something wrong but that's also possible. What would you prefer?

PeachesMLG: id say do it my way

PR #358, opened 2021-08-10
Use generics when possible

dlsf:


dlsf:
IntelliJ told me that that `instanceof` check can be replaced with a null check as soon as I added the generic so I guess it's fine

PeachesMLG:
I belive i didnt do this for a reason, the serializer woulddnt serialise the classes correctly? did you test this?

dlsf:
No, I didn't

PR #360, opened 2021-08-11
Make portals are works when nether or end islands disabled

MertUnverdi:
Wont teleport if nether or end island is enabled but the portal is on a island world and island not present


PeachesMLG:
Could u explain what this does? Im too dum to figure it out

MertUnverdi:
> Could u explain what this does? Im too dum to figure it out

just merge it its fine 👍🏻

MertUnverdi:
@dlsf could you explain it to 🍑  ?

MertUnverdi:
Are u serious ????

PR #371, opened 2021-08-23
Add permissions by commands

DoctaEnkoda:


dlsf:
You can add permissions to commands in the commands.yml, but most people don't want permissions for each command by default.

What's the purpose of this PR?

DoctaEnkoda:
The goal I had was to deactivate certain commands to redo them with the Java GUI and Bedrock edition  (Geyser+Floodgate)

dlsf:
That sounds pretty confusing and specific. Could you explain a bit more?

I don't think that this should end up in the plugin, though.

PR #374, opened 2021-08-28
Add proper subcommand system, Cleanup CommandManager code

dlsf:
I also shortened the code that registers all commands. I know that @PeachesMLG didn't want this last time but I managed to find a solution without using a third-party library that is in my opinion way better than the manual registration. I can revert that part if it's still not wanted.

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/commands/BankCommand.java
+++ b/src/main/java/com/iridium/iridiumskyblock/commands/BankCommand.java
@@ -2,9 +2,9 @@ package com.iridium.iridiumskyblock.commands;
 
 import com.iridium.iridiumcore.utils.StringUtils;
 import com.iridium.iridiumskyblock.IridiumSkyblock;
-import com.iridium.iridiumskyblock.commands.subcommands.BankGive;
-import com.iridium.iridiumskyblock.commands.subcommands.BankRemove;
-import com.iridium.iridiumskyblock.commands.subcommands.BankSet;
+import com.iridium.iridiumskyblock.commands.bank.GiveCommand;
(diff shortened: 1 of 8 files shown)


src/main/java/com/iridium/iridiumskyblock/commands/BankCommand.java
```diff
@@ -24,18 +24,20 @@
  */
 public class BankCommand extends Command {
 
-    public BankGive bankGive;
-    public BankSet bankSet;
-    public BankRemove bankRemove;
+    public GiveCommand bankGive;
```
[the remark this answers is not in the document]
dlsf: These variable names are in the config, renaming them here will mess up existing configs and confuse people.

dlsf:
I obviously tested this PR, except child commands having other children.

src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
```diff
@@ -233,16 +196,40 @@ public boolean onCommand(@NotNull CommandSender commandSender, org.bukkit.comman
...
+            || command.permission.equalsIgnoreCase("")
+            || command.permission.equalsIgnoreCase("iridiumskyblock.");
+    }
+
+    private Command findExecutingCommand(Command baseCommand, String[] args) {
+        Command executingCommand = baseCommand;
+        int currentArgument = 1;
+
+        // Check for each argument if it's a child of the previous command
+        while (args.length > currentArgument) {
```
PeachesMLG, line 224: Cant this just be a for loop? or am I being dum?

dlsf: Well, yeah

Issue #362, opened 2021-08-16
When PLAYER1 is transferred to PLAYER2 island PLAYER1 creates another island after exiting the empty island He will show the island name conflict


28451553:
When PLAYER1 is transferred to PLAYER2 island
PLAYER1 creates another island after exiting the empty island
He will show the island name conflict

PR #389, opened 2021-09-05
Island name

PeachesMLG:
Fixes https://github.com/Iridium-Development/IridiumSkyblock/issues/362 by allowing the island name to be stored as null in the database and then defaulting it to the owners name.

Only issue is how to transform this data in sqllite since ALTER apparently doesnt exist?


src/main/java/com/iridium/iridiumskyblock/commands/RenameCommand.java
```diff
@@ -66,7 +66,10 @@ public boolean execute(CommandSender sender, String[] args) {
             ));
             return false;
         }
```
dlsf, line 68:
```suggestion
        }
        
```

src/main/java/com/iridium/iridiumskyblock/commands/RenameCommand.java
```diff
@@ -66,7 +66,10 @@ public boolean execute(CommandSender sender, String[] args) {
...
-
+        if (IridiumSkyblock.getInstance().getIslandManager().getIslandByName(name).isPresent()) {
+            player.sendMessage(IridiumSkyblock.getInstance().getMessages().islandWithNameAlreadyExists.replace("%prefix%", (IridiumSkyblock.getInstance().getConfiguration()).prefix));
+            return false;
+        }
```
dlsf, line 72:
```suggestion
        }
        
```

dlsf (changes requested):

src/main/resources/patch_2_mysql.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE islands MODIFY name VARCHAR(512) NOT NULL
```
PeachesMLG, line 1: Shouldnt this be null?

dlsf: maybe

PR #390, opened 2021-09-06
Reformat & Refactor commands,  Shorten tab-completion code 

dlsf:


src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
```diff
@@ -233,4 +239,14 @@ private Command findExecutingCommand(Command baseCommand, String[] args) {
         return executingCommand;
     }
 
+    private List<String> filterTabCompletionResults(List<String> tabCompletion, String[] arguments) {
+        return tabCompletion.stream()
+            .filter(completion -> {
+                String unifiedCompletion = completion.toLowerCase();
+                String lastArgument = arguments[arguments.length - 1].toLowerCase();
+                return unifiedCompletion.startsWith(lastArgument) || unifiedCompletion.contains(lastArgument);
```
PeachesMLG, line 247: do we need to or here? if it starts with it it will also contain it?

PeachesMLG (changes requested):
I checked most of them..... ok some of them.... ok I skimmed over some of them then gave up half way through

PR #409, opened 2021-09-20
Fix bank item names being null

MertUnverdi:


dlsf:
What issue does this fix?

MertUnverdi:
> What issue does this fix?

@Faun471 has added displayName for bank items, it wont be added for preexiting bank.yml try it and see 👎🏻

PR #415, opened 2021-10-01
Add click to join button to invite command

dlsf:
Requested by a user

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/commands/InviteCommand.java
+++ b/src/main/java/com/iridium/iridiumskyblock/commands/InviteCommand.java
@@ -106,7 +106,7 @@ public class InviteCommand extends Command {
         if (offlinePlayer instanceof Player) {
             Player targetPlayer = (Player) offlinePlayer;
             TextComponent message = new TextComponent(StringUtils.color(IridiumSkyblock.getInstance().getMessages().youHaveBeenInvited));
-            message.setClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/is join " + player.getName()));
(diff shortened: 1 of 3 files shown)


src/main/java/com/iridium/iridiumskyblock/commands/InviteCommand.java
```diff
@@ -100,7 +104,11 @@ public boolean execute(CommandSender sender, String[] args) {
 
         // Send a message to the user if he is online
         if (offlinePlayer instanceof Player) {
-            ((Player) offlinePlayer).sendMessage(StringUtils.color(IridiumSkyblock.getInstance().getMessages().youHaveBeenInvited.replace("%inviter%", player.getName()).replace("%prefix%", IridiumSkyblock.getInstance().getConfiguration().prefix)));
+            Player targetPlayer = (Player) offlinePlayer;
+            TextComponent message = new TextComponent(StringUtils.color(IridiumSkyblock.getInstance().getMessages().youHaveBeenInvited));
+            message.setClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/is join " + player.getName()));
```
PeachesMLG, line 109: what if the owners are really dumb anc changed /is join to something else or changed the language or whatever, dont think we should hard code

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/commands/InviteCommand.java
dlsf: Right, I'll change that.

PeachesMLG (approved):

PR #416, opened 2021-10-01
Add more placeholders

dlsf:
New placeholders containing the names of all members, visitors, and players on an island as well as the number of visitors and players.

This was requested by a user.


MertUnverdi:
i think _players and _player_names is unnesecary but why not

MertUnverdi (approved):

PR #419, opened 2021-10-03
Add toggle for the island menu

dlsf:


src/main/java/com/iridium/iridiumskyblock/commands/CommandManager.java
```diff
@@ -107,7 +107,12 @@ public boolean onCommand(@NotNull CommandSender commandSender, org.bukkit.comman
                 User user = IridiumSkyblock.getInstance().getUserManager().getUser(player);
                 Optional<Island> island = user.getIsland();
                 if (island.isPresent()) {
-                    player.openInventory(new InventoryConfigGUI(IridiumSkyblock.getInstance().getInventories().islandMenu).getInventory());
+                    if (IridiumSkyblock.getInstance().getConfiguration().disableMainMenu) {
+                        String command = IridiumSkyblock.getInstance().getCommands().helpCommand.aliases.get(0);
```
PeachesMLG, line 111: So I realised Im being really dumb when doing stuff this way, we should just do helpCommand.execute. Idk if u wana put another pr through to do that, if not ill do it

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
```diff
@@ -38,6 +38,7 @@
     public boolean respawnOnIsland = true;
     public boolean obsidianBucket = true;
     public boolean patreonMessage = true;
+    public boolean disableIslandMenu = false;
```
PeachesMLG, line 41: also I think this should be the opposite
public boolean islandMenu = true

PeachesMLG: so instead of disabling it have the boolean be to enable it and default to true

dlsf: Yeah, I was thinking the same, no idea why I didn't do it

PR #425, opened 2021-10-03
Added some Placeholders suggested on discord channel

MertUnverdi:


src/main/java/com/iridium/iridiumskyblock/configs/Placeholders.java
```diff
@@ -26,15 +26,15 @@
     public String islandBiome = "N/A";
 
     // Island Upgrade Placeholders
-    public String islandUpgradeBlocklimitLevel = "0";
-    public String islandUpgradeMemberLevel = "0";
-    public String islandUpgradeSizeLevel = "0";
-    public String islandUpgradeGeneratorLevel = "0";
-    public String islandUpgradeWarpLevel = "0";
```
dlsf, line 33: I don't think that we should remove these options, they would pollute old configs + maybe someone wants different placeholders.

dlsf (changes requested):

src/main/java/com/iridium/iridiumskyblock/configs/Placeholders.java
MertUnverdi: i don't think no one uses this :/

PeachesMLG: agree with das, keep this in

dlsf: A use case would be setting `islandUpgradeGeneratorLevel` to `No generator` or something similar

dlsf (approved):

PR #433, opened 2021-10-04
If server has FAWE or AWE, use async task for paste

DoctaEnkoda:


PeachesMLG:
This should return a CompletableFuture instead, we dont want to teleport players to the island until the island has been pasted

DoctaEnkoda:
> This should return a CompletableFuture instead, we dont want to teleport players to the island until the island has been pasted


I just took the paste async function, the CompletableFuture was not changed.

PeachesMLG:
Dont think you quite understood what I ment so I did it myself here https://github.com/Iridium-Development/IridiumSkyblock/commit/c8b369a88224010b3169511df10e0b24a5583ced.

Thanks alot of the pr though, I really appreciate it :)

PR #368, opened 2021-08-17
New Heights 1.17+

DoctaEnkoda:
Since 1.17, Mojang has brought the new heights and Spigot has therefore, at the same time an API for it even if it is in datapack.
We can use it now!

PR #449, opened 2021-10-09
This PR follows the PR: IridiumSkyblock#368

Euphillya:
This PR follows the PR: #368 

Since we can put water below layer 0, we can also be trying to make an underwater island, the concern is that if we put a dome, it is filled with water and it is not playable! This PR allows the blow to ignore or not the installation of air block.


MertUnverdi:
i don't think its matter but 👍🏻 

> i don't think its matter but 👍🏻 also the default boolean is false so pre exiting configs will be false by default its really important cuz can be there too much big schematics.. add a way to making defaults as true

sorry i didn't saw it wasn't primitive boolean so you can check "ignoreAirBlock" is null for making defaults as true (like how i did for islandHeight thing)

DoctaEnkoda:
No worries, I just did it;)

PR #464, opened 2021-10-15
Adding back buttons

MBDOFF:
Adding back buttons to almost all GUI's.


PeachesMLG:
Replaced By https://github.com/Iridium-Development/IridiumSkyblock/pull/497

PR #491, opened 2021-11-07
Fix mushroom schematic generate nothing.

DreamVoid:
Since the 3.1.2 version supports WorldEdit, when the player selects "mushroom" to generate an empty island, no blocks will be generated.
Besides, nether and end will be generated normally


MertUnverdi:
its @PeachesMLG's fault blame him

dlsf:
What schematic are you using now? The old mushroom schematic?

PR #493, opened 2021-11-12
Add Back Buttons each gu

MertUnverdi:


src/main/java/com/iridium/iridiumskyblock/BackButton.java
```diff
@@ -0,0 +1,14 @@
...
+import com.iridium.iridiumcore.Item;
+import lombok.AllArgsConstructor;
+import lombok.NoArgsConstructor;
+
+@AllArgsConstructor
+@NoArgsConstructor
+public class BackButton {
+
+    public boolean enabled;
+    public String command;
```
PeachesMLG, line 12: I think this should be an Inventory instead of a command, then we can optionally paste the Back button if that inventory is null or not

src/main/java/com/iridium/iridiumskyblock/IridiumSkyblock.java
```diff
@@ -513,6 +516,34 @@ public void loadConfigs() {
         Bukkit.getPluginManager().callEvent(reloadEvent);
     }
 
+    private void updateBackButtons() {
+        BackButton backButton22Is = new BackButton(true, "is", new Item(XMaterial.NETHER_STAR, 22, 1, "&c&lBack", Collections.emptyList())); //Used for making inline if statements
```
PeachesMLG, line 520: just have a BackButton that is relative to the inventory size, e.g. have a slot of -5 which will put it in the slot inventorysize-5

PeachesMLG (changes requested):

PeachesMLG:
Ima do this myself, this seems super messy

PeachesMLG:
Replaced by https://github.com/Iridium-Development/IridiumSkyblock/pull/497

PR #495, opened 2021-11-14
Make default island colours based on Schematics

MertUnverdi:


PeachesMLG:
I dont really see a need for this?

dlsf:
Looks really unnecessary. We should add the schematic to the IslandCreateEvent though, if that's possible.

MertUnverdi:
but, why not??? also this suggested on suggestions channel

MertUnverdi:
it still a feature so can be added, if wont, close the pr i hate to wait "(

Issue #482, opened 2021-10-31
Bungeecord breaks /is visit


dlsf:
**Describe the bug**
You can't teleport to islands of offline players when using Bungeecord.

**To Reproduce**
1. Run `/is visit <name of an offline player>`
2. "No island found for that player"

**Expected behavior**
You should be teleported to the island.

**Server Version and Plugin Version**
IridiumSkyblock version: 3.1.4
Server version: git-Paper-349 (MC: 1.17.1)

**Additional context**
Might be related to Bungeecord, but it is set up correctly (online mode Bungeecord, offline mode server, `settings.bungeecord=true` in the `spigot.yml`).

GavvyDizzle:
My server uses Bungee and I don't have this issue

ajh123:
If we have this plugin on a lobby server, people cannot ```/is visit``` to their island on another proxied server e.g. skyblock

PR #496, opened 2021-11-14
Add offline player support to bank commands

dlsf:


MertUnverdi (changes requested):
Make it for VisitCommand tho

dlsf:
> Make it for VisitCommand tho

I already mentioned this a few times now, but [we already do this in the visit command](https://github.com/Iridium-Development/IridiumSkyblock/blob/master/src/main/java/com/iridium/iridiumskyblock/commands/VisitCommand.java#L47).

#482 is most likely related to BungeeCord.

PR #503, opened 2021-11-15
Improve event api

dlsf:
- Add documentation & @NotNull annotations
- Add ShopPurchaseEvent
- Add ShopSellEvent
- Add IslandRenameEvent
- Implement & Extend IslandSettingChangeEvent
- Enhanced IslandCreateEvent

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/api/IslandCreateEvent.java
+++ b/src/main/java/com/iridium/iridiumskyblock/api/IslandCreateEvent.java
@@ -1,22 +1,33 @@
 package com.iridium.iridiumskyblock.api;
 
+import com.iridium.iridiumskyblock.configs.Schematics;
 import com.iridium.iridiumskyblock.database.User;
 import lombok.Getter;
+import lombok.Setter;
 import org.bukkit.event.Cancellable;
 import org.bukkit.event.Event;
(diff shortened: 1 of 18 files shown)


src/main/java/com/iridium/iridiumskyblock/api/IslandCreateEvent.java
```diff
@@ -1,22 +1,33 @@
...
+ * Called before an Island has been created.
+ */
 @Getter
 public class IslandCreateEvent extends Event implements Cancellable {
+
     private static final HandlerList handlers = new HandlerList();
     private boolean cancelled;
-    private final String islandName;
-    private final User user;
+    @NotNull private final String islandName;
```
PeachesMLG, line 20: dis should be nullable?

src/main/java/com/iridium/iridiumskyblock/api/IslandCreateEvent.java
```diff
@@ -1,22 +1,33 @@
...
+    @NotNull private final User user;
+
+    @NotNull
```
PeachesMLG: put the @NotNull on the same line with the @Setter above it

PeachesMLG (changes requested):

PR #504, opened 2021-11-16
Fix command invocations in GUIs

dlsf:


PeachesMLG:
This has already been fixed?

PR #508, opened 2021-11-20
Fixed Bug Crops Mine Mission

DoctaEnkoda:
Fixed a bug in which if we break crops in missions, it does not check if it is mature.


MertUnverdi (approved):

dlsf (approved):

PeachesMLG:
For the record, thats what we added the "GROW" Mission for because if its something like cactus they can just place/break over and over again where as grow is when it actually grows naturally

PR #509, opened 2021-11-20
Fix Possible NullPointerException

DoctaEnkoda:
Fix Possible NullPointerException With Essentials


dlsf (approved):

src/main/java/com/iridium/iridiumskyblock/utils/PlayerUtils.java
```diff
@@ -95,10 +95,12 @@ public static void teleportSpawn(Player player) {
         if (Bukkit.getPluginManager().isPluginEnabled("EssentialsSpawn")) {
```
PeachesMLG, line 95: Dont we already check for that here?

DoctaEnkoda: API says the answer can be null, better avoid a problem in the future

DoctaEnkoda: Crash Plugin, Disabled with an issue...

PeachesMLG: ok, well we should remove the if statement above and replace it with your ones then, seems kinda dum to have both?

PeachesMLG (changes requested):
Discussed changes

PeachesMLG:
Fixed Issue with if statements

PR #516, opened 2021-11-25
Fix permissions GUI pagination again

dlsf:


dlsf:
This method did something else at first, that's why there was a '>'

PeachesMLG (approved):

PR #519, opened 2021-11-29
Fix mobs targeting visitors when they shouldn't

dlsf:


src/main/java/com/iridium/iridiumskyblock/listeners/EntityTargetListener.java
```diff
@@ -34,6 +34,9 @@ public void onEntityTargetEntity(EntityTargetLivingEntityEvent event) {
         Optional<IslandTrusted> targetTrusted = IridiumSkyblock.getInstance().getIslandManager().getIslandTrusted(island.get(), targetUser);
         if (island.get().equals(targetUser.getIsland().orElse(null)) || targetTrusted.isPresent()) return;
 
+        // Player is a visitor and mobsVisitorTargeting is disabled
+        event.setCancelled(true);
```
PeachesMLG, line 38: Unless im dumb, we shouldnt cancel here? if we cancel the event then the mobs wont target anyone?

PR #526, opened 2021-12-02
Fix Fishing Missions

GavvyDizzle:
The old code did not properly pass the material of the item that was caught.


src/main/java/com/iridium/iridiumskyblock/listeners/PlayerFishListener.java
```diff
@@ -24,7 +25,7 @@ public void monitorPlayerFish(PlayerFishEvent event) {
         User user = IridiumSkyblock.getInstance().getUserManager().getUser(event.getPlayer());
         Optional<Island> island = user.getIsland();
 
-        island.ifPresent(value -> IridiumSkyblock.getInstance().getMissionManager().handleMissionUpdates(value, "FISH", caughtEntity.getType().name(), 1));
+        island.ifPresent(value -> IridiumSkyblock.getInstance().getMissionManager().handleMissionUpdates(value, "FISH", ((Item) caughtEntity).getItemStack().getType().toString(), 1));
```
dlsf, line 28: Use Material#name(), it's just good practice 
```suggestion
        island.ifPresent(value -> IridiumSkyblock.getInstance().getMissionManager().handleMissionUpdates(value, "FISH", ((Item) caughtEntity).getItemStack().getType().name(), 1));
```

dlsf (changes requested):

PeachesMLG (commented):
Is Caught Entity always an item? I belive in some plugins they modify it so you can fish up tnt and stuff, should we add an instance check?

GavvyDizzle:
PlayerFishEvent.State.CAUGHT_FISH

When a player has successfully caught a fish and is reeling it in. In this instance, a "fish" is any item retrieved from water as a result of fishing, ie an item, but not necessarily a fish.

I think the check to make sure it is a CAUGHT_FISH guarentees that it is an item.

dlsf:
> Is Caught Entity always an item? I belive in some plugins they modify it so you can fish up tnt and stuff, should we add an instance check?

There is no way to alter the item that has been fished. I've written a fishing plugin myself, and I ended up having to re-code the entire Minecraft reel-in system just to change the item / entity that you fish. If I didn't overlook anything back then, the item will always be a fish, or else the event will have to be cancelled by the fishing plugin.

dlsf (approved):

PR #530, opened 2021-12-09
Made IslandLevelEquation configurable

PeachesMLG:
Thoughts on how we do the getExperienceRequired method?

We could just do a brute force approach but that seems dum

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
+++ b/src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
@@ -31,6 +31,7 @@ public class Configuration {
     public String dateTimeFormat = "EEEE, MMMM dd HH:mm:ss";
     public String spawnWorldName = "world";
     public String defaultCommandSyntax = "/is %command%";
+    public String islandLevelEquation = "abs(cbrt($1+1))";
 
     public boolean enableCheckVersion = true;
     public boolean defaultIslandPublic = true;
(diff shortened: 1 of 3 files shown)


src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
```diff
@@ -31,6 +31,7 @@
     public String dateTimeFormat = "EEEE, MMMM dd HH:mm:ss";
     public String spawnWorldName = "world";
     public String defaultCommandSyntax = "/is %command%";
+    public String islandLevelEquation = "abs(cbrt($1+1))";
```
dlsf, line 34: Why is there an abs() call? All cube roots are positive

src/main/java/com/iridium/iridiumskyblock/database/Island.java
```diff
@@ -112,6 +115,7 @@ public int getLevel() {
      * @return The experience required to reach this level
      */
     private int getExperienceRequired(int level) {
+        //TODO
         return -1 + (level * level * level);
```
dlsf, line 119: Maybe we can create an equation like `level=cbrt(x)` where level is defined, and we can just solve for x? I don't know if Crunch can do that, though.

src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
PeachesMLG: :shrug: It was in my first level equation code and I figured there was probabbly a reason, and if it is useless, it does litrally nothing so meh

PR #531, opened 2021-12-10
Added createIslandOnHome

Faun471:


src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
```diff
@@ -42,6 +42,7 @@
     public boolean patreonMessage = true;
     public boolean islandMenu = true;
     public boolean backButtons = true;
+    public boolean createIslandOnHome = false;
```
PeachesMLG, line 45: default this to true

dlsf:
Maybe this should be a setting in the commands.yml? This would make the configuration.yml less cluttered.

When Peaches agrees, you just have to add a public boolean to the HomeCommand class.

Issue #489, opened 2021-11-04
Disappearance of the islands


DoctaEnkoda:
**Describe the bug**
On my 3 servers, I have over 3000 islands according to the Islands table but from time to time, I have players who no longer own their island for some unexplained reason. It's just the island_id value that disappears in users table, I manage to give them their island manually via a quick code. 

**To Reproduce**
Peut pas reproduire car c'est inexpliquable

**Server and Plugin Version**
Version : 
```
[15:45:59 INFO]: Checking version, please wait...
[15:46:00 INFO]: The_black_shade issued server command: /is go
[15:46:00 INFO]: This server is running Purpur version git-Purpur-1419 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: b72753c on ver/1.17.1)
[15:46:00 INFO]: You are 4 version(s) behind
Download the new version at: https://purpur.pl3x.net/downloads/
```
/is about
```
[15:46:41 INFO]: Plugin Name: IridiumSkyblock
[15:46:41 INFO]: Plugin Version: 3.1.5
[15:46:41 INFO]: Plugin Author: Peaches_MLG
[15:46:41 INFO]: Plugin Contributors: das_, SlashRemix
[15:46:41 INFO]: Plugin Donations: www.patreon.com/Peaches_MLG
```

DoctaEnkoda:
I will try to explain at best because even I do not understand:

Occasionally, for no apparent reason, mostly island OWNERS lose their island_id / Role / Playtime.
The island still exists but it's really the user table that has a problem.

The users table is currently: 12266 rows.
From time to time, it comes back but it is quite rare.

I had to do this house code to give them back but it's clearly not clean:
```Java
try{
                Optional<User> user = IridiumSkyblock.getInstance().getUserManager().getUserByUUID(UUID.fromString(args[1]));
                User user1 = user.get();
                Island island = IridiumSkyblock.getInstance().getIslandManager().getIslandById(Integer.parseInt(args[0])).get();
                user1.setIsland(island);
                user1.setIslandRank(IslandRank.OWNER);
            } catch (Exception e) {
                sender.sendMessage(ChatColor.RED + e.getMessage());
            }
```

PeachesMLG:
See if this is fixed in 3.2.0

DoctaEnkoda:
Following this bug, during our tests, we saw that if a person who lost their island tries to recreate, the plugin crashes with its logs:
[...]

PR #537, opened 2021-12-18
Patch Clear Island

DoctaEnkoda:
The code that I submit and made in Java 16, will have to modify the functions in Java 8
https://github.com/Iridium-Development/IridiumSkyblock/issues/489

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/TableManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/TableManager.java
@@ -60,8 +60,8 @@ public class TableManager<T extends DatabaseObject, S> {
         }
     }
     
-    public void saveHashMap(LinkedHashMap<?, T> uuidtLinkedHashMap) {
-        List<T> tList = new LinkedList<>(uuidtLinkedHashMap.values());
+    public void saveHashMap(LinkedHashMap<?, T> uuidLinkedHashMap) {
+        List<T> tList = new LinkedList<>(uuidLinkedHashMap.values());
         int sizeList = tList.size();
         try {
             for (int i = 0; i < sizeList; i++) {
--- a/src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/TableManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/TableManager.java
@@ -55,7 +55,7 @@ public class TableManager<T extends DatabaseObject, S> {
     }
     
     public void saveHashMap(LinkedHashMap<?, T> uuidtLinkedHashMap) {
-        List<T> tList = uuidtLinkedHashMap.values().stream().toList();
+        List<T> tList = new LinkedList<>(uuidtLinkedHashMap.values());
         int sizeList = tList.size();
         try {
             for (int i = 0; i < sizeList; i++) {
--- a/src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/TableManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/TableManager.java
@@ -63,6 +63,7 @@ public class TableManager<T extends DatabaseObject, S> {
     public void saveHashMap(LinkedHashMap<?, T> uuidLinkedHashMap) {
         this.saveHashMap(uuidLinkedHashMap, true);
     }
+    
     public void saveHashMap(LinkedHashMap<?, T> uuidLinkedHashMap, boolean useChanged) {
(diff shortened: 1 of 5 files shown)


PeachesMLG:
Looks like the builds failing

src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/UserTableManager.java
```diff
@@ -14,49 +14,49 @@
...
-     * Puts the user in the correct order of userIslandIndex
+     * Puts the user in userIslandMap
      *
      * @param user The specified User
      */
     public void resortIsland(User user) {
-        userIslandIndex.remove(user);
-        int islandIndex = Collections.binarySearch(userIslandIndex, user, Comparator.comparing(u -> u.getIsland().map(Island::getId).orElse(0)));
-        userIslandIndex.add(islandIndex < 0 ? -(islandIndex + 1) : islandIndex, user);
+        if (userIslandMap.get(user.getUuid()) != null) {
```
PeachesMLG, line 44: This method was origionally used to sort the User's via the IslandID, so we can get a list of all members in an island.

DoctaEnkoda: But now, its not list, its linkedhashmap

DoctaEnkoda:
He has fail because my code use function Java 11, I don't have much time to convert it to java 8.
In addition, moving to Java 11 will only do more good to continue developing on it.

dlsf:
But we would lose 25% of our users, which is a lot: https://bstats.org/plugin/bukkit/IridiumSkyblock/5825

DoctaEnkoda:
Build Java 8 OK

PeachesMLG:
Have you tested this?

Im currently adding unit tests to Iridium, and im not really confident this will be bug free, ima add unit tests and then see if they all pass with this version.

DoctaEnkoda:
With this patch, i'm 0 lost island on my server, with actual code, i'm 20 lost island / 4000 players per day

PeachesMLG:
sorry bout the delay, ill take a look at this tmr

src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/TableManager.java
```diff
@@ -62,6 +59,30 @@ public void save() {
             exception.printStackTrace();
         }
     }
+    
+    public void saveHashMap(LinkedHashMap<?, T> uuidLinkedHashMap) {
+        this.saveHashMap(uuidLinkedHashMap, true);
+    }
```
dlsf, line 65:
```suggestion
    }
    
```

src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/TableManager.java
```diff
@@ -62,6 +59,30 @@ public void save() {
...
+    public void saveHashMap(LinkedHashMap<?, T> uuidLinkedHashMap, boolean useChanged) {
+        List<T> tList = new LinkedList<>(uuidLinkedHashMap.values());
```
dlsf, line 67: What's the point in the `t` prefix here?

src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/TableManager.java
```diff
@@ -62,6 +59,30 @@ public void save() {
...
+        int sizeList = tList.size();
+        try {
+            for (int i = 0; i < sizeList; i++) {
+                T t = tList.get(i);
+                if (!useChanged) { // all save
```
dlsf, line 72: What exactly is the use case for this toggle?

dlsf (changes requested):

src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/TableManager.java
DoctaEnkoda: For save user data

src/main/java/com/iridium/iridiumskyblock/managers/tablemanagers/TableManager.java
DoctaEnkoda: Just value is T

DoctaEnkoda:
I will include a new PR after this patch because, another error that was causing the plugin to crash was found.

Hyronymos1337:
any update on this?

DoctaEnkoda:
closed because does not intend to fix it

Issue #544, opened 2021-12-30
Stacked blocks(RoseStacker) are not counted when there is no one on the island. Bug Report


iliakravets677:
**Describe the bug**
Stacked blocks(RoseStacker) are not counted when there is no one on the island.

**To Reproduce**
If you create Stacked blocks using RoseStacker on your island, and leave the island, it will count one block, but Stacked blocks will not

**Expected behavior**
So that the block rating is always correct.
rosestacker 1.4.1
![image](https://user-images.githubusercontent.com/56079131/147772504-4217dbab-8af6-4d0a-9c87-38ceef95e549.png)

**Server and Plugin Version**
Copy and paste the results of `/version` and `/is about`

**Screenshots**
![image](https://user-images.githubusercontent.com/56079131/147772538-cfb1e3dd-9173-433f-9a85-478b107b0a9c.png)
![image](https://user-images.githubusercontent.com/56079131/147772546-a635f10d-ff5a-460b-b0a5-197527df0e7f.png)
![image](https://user-images.githubusercontent.com/56079131/147772553-05253f95-9296-49bf-84a1-430af1de7c78.png)

**Logs**
no errors

**Additional context**

PeachesMLG:
So it looks like the code we are using to get the StackedBlocks only returns the stackedblocks in loaded chunks

iliakravets677:
> So it looks like the code we are using to get the StackedBlocks only returns the stackedblocks in loaded chunks

it can be fixed?

dlsf:
Yeah, we are working on it

iliakravets677:
> Yeah, we are working on it

also check the work with loading when the owner of the island is in another world

PR #545, opened 2021-12-31
Fixed Issues with RoseStacker

PeachesMLG:
Fixed https://github.com/Iridium-Development/IridiumSkyblock/issues/544


src/main/resources/patch_4.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE island_blocks ADD COLUMN extra_amount INTEGER PRECISION DEFAULT 0.00;
+ALTER TABLE island_spawners ADD COLUMN extra_amount INTEGER PRECISION DEFAULT 0.00;
```
dlsf, line 2: I don't think that these SQL statements work with MySQL. Ran a checker I found online, and it shows errors.

dlsf (changes requested):

PR #546, opened 2022-01-01
Fix fire ticks harming visitors

dlsf:


PeachesMLG:
Hold up, whats FIRE_TICK and how is it different from FIRE

dlsf:
> Hold up, whats FIRE_TICK and how is it different from FIRE

My guess based on the bug report is that FIRE is used when the player is standing inside a block of fire, while FIRE_TICK is used when he is just on fire but not in lava or in a block of fire.

PR #551, opened 2022-01-02
Improve RandomAccessList & Add unit tests

dlsf:


src/main/java/com/iridium/iridiumskyblock/utils/RandomAccessList.java
```diff
@@ -14,42 +16,96 @@
...
-     * @param possibility The possibility of this element being selected.
+     * @param element The element which should be added
+     * @param possibility The possibility of this element being selected
      */
-    public void add(E e, int possibility) {
-        for (int i = 0; i < possibility; i++) {
-            underlyingList.add(e);
-        }
+    public void add(E element, int possibility) {
+        elements.add(new RandomAccessElement<>(element, new IntRange(size, size + possibility - 1)));
```
PeachesMLG, line 30: Wont this lead to an infinite loop?

src/main/java/com/iridium/iridiumskyblock/utils/RandomAccessList.java
```diff
@@ -14,42 +16,96 @@
  */
 public class RandomAccessList<E> {
 
-    private final List<E> underlyingList = new ArrayList<>();
+    private final List<RandomAccessElement<E>> elements = new ArrayList<>();
```
PeachesMLG, line 19: Why do we have a List of RandomAccessElements? I dont really like this, Feel like we should have a Hashmap instead?

PeachesMLG (changes requested):
I dont really Like this

PeachesMLG:
Thoughts on https://github.com/Iridium-Development/IridiumSkyblock/pull/552 as an alternitive?

PR #560, opened 2022-01-10
Add CurrentOffset command

dlsf:
It would probably be better to have it as a subcommand for the admin command, but I believe that this is a good (temporary) solution.


PeachesMLG:
What would this be used for? Setting default home locations?
Maybe we should add a new command like `isadmin setDefaultHome` or something

Also im planning on adding alot of unit tests to IridiumSkyblock, so if u add new things if you could add unit tests to the commands that would be great

PR #561, opened 2022-01-14
Fix issue Async task on entities

DoctaEnkoda:
I found the bug by putting a catch on IslandManager#recalculateIsland(Island)

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
@@ -866,15 +866,27 @@ public class IslandManager {
                 }
             }
         });
-        chunks.forEach(chunk -> {
-            for (BlockState blockState : chunk.getTileEntities()) {
-                if (!(blockState instanceof CreatureSpawner)) continue;
-                if (!island.isInIsland(blockState.getLocation())) continue;
-                CreatureSpawner creatureSpawner = (CreatureSpawner) blockState;
-                IslandSpawners islandSpawners = IridiumSkyblock.getInstance().getIslandManager().getIslandSpawners(island, creatureSpawner.getSpawnedType());
-                islandSpawners.setAmount(islandSpawners.getAmount() + 1);
-            }
-        });
+        if (Bukkit.isPrimaryThread() == false) {
+            Bukkit.getScheduler().runTask(IridiumSkyblock.getInstance(), () -> {
+                chunks.forEach(chunk -> {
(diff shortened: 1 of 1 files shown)


src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -866,15 +866,29 @@ private void recalculateIsland(@NotNull Island island, @NotNull List<Chunk> chun
...
-        chunks.forEach(chunk -> {
-            for (BlockState blockState : chunk.getTileEntities()) {
-                if (!(blockState instanceof CreatureSpawner)) continue;
-                if (!island.isInIsland(blockState.getLocation())) continue;
-                CreatureSpawner creatureSpawner = (CreatureSpawner) blockState;
-                IslandSpawners islandSpawners = IridiumSkyblock.getInstance().getIslandManager().getIslandSpawners(island, creatureSpawner.getSpawnedType());
-                islandSpawners.setAmount(islandSpawners.getAmount() + 1);
-            }
-        });
+        if (Bukkit.isPrimaryThread() == false) {
```
PeachesMLG, line 869: Change this to
```java
if(Bukkit.isPrimaryThread()){

}else{
Bukkit.getScheduler().runTask(IridiumSkyblock.getInstance(), this::recalculateIsland);
}
```
since its kinda dum to repeat the same code twice

PeachesMLG:
Also what bug is this supposed to be fixing?

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
DoctaEnkoda: No need to go through everything synchronously because it is only calling the entity and it risks duplicating the data because it would be necessary to re-reset the cache.

DoctaEnkoda: I'm create a function for it

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -866,6 +866,14 @@ private void recalculateIsland(@NotNull Island island, @NotNull List<Chunk> chun
                 }
             }
         });
```
dlsf, line 868:
```suggestion
        });
        
```

dlsf:
Did you test this fix?

DoctaEnkoda:
Yes

dlsf (approved):
Looks fine

PR #563, opened 2022-01-14
Disable Task start if multiverse is not on server

DoctaEnkoda:


dlsf (approved):
I mean... the difference is pretty much non-existent, but I guess it isn't necessarily bad either.

PR #566, opened 2022-01-14
Fix right-clicking depositing two crystals at once

dlsf:
I just tested it but forgot what the issue is. You either deposit two crystals when you do or don't click on a block. There is an explanation in the code. Spigot is dum


PeachesMLG:
iirc there was an event .gethand or something like that, would this not be better?

dlsf:
Right, I confused left hand and left click

PR #568, opened 2022-01-16
Fix issue with +54 rewards

DoctaEnkoda:


src/main/java/com/iridium/iridiumskyblock/gui/IslandRewardsGUI.java
```diff
@@ -27,11 +27,14 @@ public void addContent(Inventory inventory) {
...
-
+        
+        int maxSize = inventory.getSize() - 1; // slot 0
+        
         AtomicInteger atomicInteger = new AtomicInteger(0);
-        IridiumSkyblock.getInstance().getDatabaseManager().getIslandRewardTableManager().getEntries(getIsland()).forEach(islandReward ->
-                inventory.setItem(atomicInteger.getAndIncrement(), ItemStackUtils.makeItem(islandReward.getReward().item, placeholders))
-        );
+        for (IslandReward islandReward : IridiumSkyblock.getInstance().getDatabaseManager().getIslandRewardTableManager().getEntries(getIsland())) {
+            if (atomicInteger.get() > maxSize) continue;
```
dlsf, line 35: You can just break, can't you?

DoctaEnkoda: yes mb

dlsf (approved):

PeachesMLG:
Good spot, if we are adding this, I think we should add next page / previous page buttons aswell

PeachesMLG (changes requested):
Good spot, if we are adding this, I think we should add next page / previous page buttons aswell

dlsf:
I don't really feel like that's needed, rewards just appear after you claim them. Why would you want to claim a specific reward?

PeachesMLG:
idk, people are weird

dlsf:
@DoctaEnkoda are you planning on adding pagination support or should someone else do it?

DoctaEnkoda:
Frankly, the pagination is useless because we only recover. Why weigh down something that doesn't need it.

PR #569, opened 2022-01-17
Change deprecated method

DoctaEnkoda:


dlsf:
Are you sure that this is the only part of the codebase where we use this?

DoctaEnkoda:
I don't see elsewhere

dlsf (approved):
I can confirm that it's the only place where we use that deprecated method

PR #577, opened 2022-01-22
Optimize something

dlsf:
Very gut PR


PeachesMLG:
Does this serve any purpose? Why only add a return there?

dlsf:
Look at the context

PR #586, opened 2022-01-25
Fix void issue (again)

MertUnverdi:
I edited it on phone could be there tab errors

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
@@ -134,6 +134,13 @@ public class IslandManager {
         boolean trusted = getIslandTrusted(island, user).isPresent();
         boolean inIsland = user.getIsland().map(Island::getId).orElse(0) == island.getId();
         if (!island.isVisitable() && !inIsland && !trusted && !user.isBypassing()) {
+            if (island.isInIsland(player.getLocation())) {
+                if (user.getIsland().isPresent()) {
+                    teleportHome(player, user.getIsland().get(), delay);
+                } else {
+                    PlayerUtils.teleportSpawn(player);
+                }
+            }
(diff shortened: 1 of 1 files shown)


dlsf (approved):
Looks good, although I don't know if there's a shorter way

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -134,6 +134,13 @@ public void teleportHome(@NotNull Player player, @NotNull Island island, int del
         boolean trusted = getIslandTrusted(island, user).isPresent();
         boolean inIsland = user.getIsland().map(Island::getId).orElse(0) == island.getId();
         if (!island.isVisitable() && !inIsland && !trusted && !user.isBypassing()) {
+            if (island.isInIsland(player.getLocation())) {
+                if (user.getIsland().isPresent()) {
+                    teleportHome(player, user.getIsland().get(), delay);
```
PeachesMLG, line 139: Well if the island is private, we dont want to send them to the island home

PeachesMLG: Also this will result in an infinite loop?

dlsf: No, this will send the player to his own island, not the island he is currently trying to visit

PeachesMLG (changes requested):
Seems like this should be specific for PlayerMoeveListener, i dont want this to happen if I do /is visit for example

MertUnverdi:
it just teleports if you cannot visit this island and already in this island

PeachesMLG:
> 

Yeah, good point. still think its better to move this logic into the playerMoveListener, since we only care if that happens there. Dont really want to crowd this method with stuff thats not as important

PR #595, opened 2022-02-04
Add option to disable setting/permission modifications

dlsf:


PeachesMLG:
not sure what the usecase of this would be?

dlsf:
For example, you don't want that users can block mob spawning or change the time?

dlsf:
And users of our plugin actually requested it, so there appears to be a use case

PR #596, opened 2022-02-04
Add command to delete other islands

dlsf:
I don't really like this solution, but I guess it's alright for now

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/commands/DeleteCommand.java
+++ b/src/main/java/com/iridium/iridiumskyblock/commands/DeleteCommand.java
@@ -38,8 +38,8 @@ public class DeleteCommand extends Command {
      */
     @Override
     public boolean execute(CommandSender sender, String[] args) {
-        if (args.length == 1) {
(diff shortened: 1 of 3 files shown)


dlsf:
This is not ready yet

src/main/java/com/iridium/iridiumskyblock/commands/DeleteCommand.java
```diff
@@ -54,6 +67,17 @@ public boolean execute(CommandSender sender, String[] args) {
         return false;
     }
 
+    private void deleteOtherIsland(CommandSender commandSender, String playerName) {
```
PeachesMLG, line 70: not sure if I like this being a seperate method

dlsf: Sure, I can inline it, but imo it's much more readable this way

dlsf: It would look like this: https://pasty.lus.pm/UuHIyP

PR #601, opened 2022-02-07
Add weight to the mission slots

dlsf:
**How this works**
When you configure a mission, you can now set the slot to numbers other than 0. This will determine the weight of this mission. The selection is still random, but the missions that have been selected will be ordered according to their weight, starting with the lowest slot. This will allow users to prevent missions from switching slots when they don't want them to.


PeachesMLG:
oh, this is for daily missions. :thinking: I still dont really like it, and dont really see why people want this. Can we make a seperate field called weight or priority or smthn?

dlsf:
> oh, this is for daily missions. thinking I still dont really like it, and dont really see why people want this. Can we make a seperate field called weight or priority or smthn?

I thought about this, but 
1) it would make the configs even longer
2) it would break existing configs, wouldn't it?

If we already have a slot, just use it. It actually works like slots are supposed to work, higher number = in a later slot

PR #614, opened 2022-02-14
Optimization Position Verification.

DoctaEnkoda:


dlsf:
When exactly is the world null? What does this improve?

DoctaEnkoda:
He is null when check : Island#isInIsland()
And call when : 
- Island#isInIsland() + screen
- IslandManager#recalculateIsland()

With 100 players on my server, we lost 8% tick with calcul function and it always call without player too.
image

dlsf:
Did you test the performance impact of your change, by any chance?

DoctaEnkoda:
Like I said, I was practically at 8% tick minimum on 100 players, especially on entity damage which was infamous on the mspt.

I can't redo spark because I'm using this modified version but yeah, went from 8% to 0.7% tick, that's pretty huge

src/main/java/com/iridium/iridiumskyblock/database/Island.java
```diff
@@ -345,6 +349,12 @@ public Location getCenter(World world) {
      * @return The Location of the first corner point
      */
     public Location getPos1(World world) {
+        if (world == null) {
+            return locationCachePos1.getCache(() -> {
+                double size = getSize() / 2.00;
+                return getCenter(null).subtract(new Location(null, size, 0, size));
+            });
+        }
```
dlsf, line 357:
```suggestion
        }
        
```

src/main/java/com/iridium/iridiumskyblock/database/Island.java
```diff
@@ -357,6 +367,12 @@ public Location getPos1(World world) {
      * @return The Location of the second corner point
      */
     public Location getPos2(World world) {
+        if (world == null) {
+            return locationCachePos2.getCache(() -> {
+                double size = getSize() / 2.00;
+                return getCenter(null).add(new Location(null, size, 0, size));
+            });
+        }
```
dlsf, line 375:
```suggestion
        }
        
```

src/main/java/com/iridium/iridiumskyblock/database/Island.java
```diff
@@ -72,6 +72,10 @@
 
     // Cache
     private Integer size;
+    
+    // Cache Position Island Reset every 5 seconds
+    private final Cache<Location> locationCachePos1 = new Cache<>(5000);
```
dlsf, line 77: I don't know if `pos1Cache` or `pos1LocationCache` would be better, this sounds a bit weird, but you don't have to change it if you don't want to

dlsf (changes requested):
The rest looks fine

PR #616, opened 2022-02-15
[CI-SKIP] Fix typo

DoctaEnkoda:


dlsf (commented):
This wasn't a typo but intentional. 

If you look at the official Javadocs from Oracle, you will notice that they are doing the same. In fact, they are even doing it everywhere except in the description, at least in this class: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#matches(java.lang.String)

PR #617, opened 2022-02-17
Improve speed of ocean generator recalculation

dlsf:


src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -856,14 +857,17 @@ public void recalculateIsland(@NotNull Island island) {
...
+
             for (int x = 0; x < 16; x++) {
                 for (int z = 0; z < 16; z++) {
                     if (island.isInIsland(x + (chunk.getX() * 16), z + (chunk.getZ() * 16))) {
                         final int maxy = Math.min(maxHeight, chunk.getHighestBlockYAt(x, z));
                         for (int y = LocationUtils.getMinHeight(world); y <= maxy; y++) {
                             XMaterial material = XMaterial.matchXMaterial(chunk.getBlockType(x, y, z));
-                            if (material.equals(XMaterial.AIR)) continue;
+                            if (material == XMaterial.AIR) continue;
+                            if (oceanWorld && material == IridiumSkyblock.getInstance().getChunkGenerator().getMainMaterial(world)) return;
```
PeachesMLG, line 870: Why are we specifying if its oceanWorld here, and why are we returning?

dlsf: We use the oceanWorld boolean here to speed up the recalculation. Checking that often if the material is in the set (or list?) will probably be pretty bad. But yeah, the return shouldn't be there

dlsf: And we should ignore air in every world

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -856,14 +856,17 @@ public void recalculateIsland(@NotNull Island island) {
...
             int maxHeight = world == null ? 255 : world.getMaxHeight() - 1;
+
             for (int x = 0; x < 16; x++) {
                 for (int z = 0; z < 16; z++) {
                     if (island.isInIsland(x + (chunk.getX() * 16), z + (chunk.getZ() * 16))) {
                         final int maxy = Math.min(maxHeight, chunk.getHighestBlockYAt(x, z));
                         for (int y = LocationUtils.getMinHeight(world); y <= maxy; y++) {
                             XMaterial material = XMaterial.matchXMaterial(chunk.getBlockType(x, y, z));
-                            if (material.equals(XMaterial.AIR)) continue;
+                            if (material == XMaterial.AIR) continue;
```
PeachesMLG, line 868: Is this a bug? think this line should be removed no?

What if its an ocean world and somone decided to drain some of the ocean, or build a base underwater or whatever. we want to regenerate that part back to water/lava no?

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
dlsf: This is the island recalculation method and has nothing to do with the world generator. 

Took me a few minutes to figure it out, tho xd

PeachesMLG: oh thought it was regenerate method mb

src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
```diff
@@ -319,6 +320,16 @@ public boolean isIslandEnd(@NotNull World world) {
         return iridiumSkyblock.getIslandManager().isIslandEnd(world);
     }
 
+    /**
+     * Returns which chunk generator IridiumSkyblock is using for world generation.
+     *
+     * @return the generator IridiumSkyblock uses
+     * @since 3.2.6
```
PeachesMLG, line 327:
```suggestion
     * @since 3.2.7
```

PR #618, opened 2022-02-17
Fix island name length

daspositioned:
767 bytes is the maximum amount of bytes for a string in some MySQL configurations, and because one character can take up to 4 bytes with utf8mb4 encoding (which is the default, apparently?), we should use a length of 190 (767/4 = 191.75, about 190).


dlsf:
https://pastebin.com/8qBnbkzA this is what the error looks like

dlsf:
I actually don't know if this fix is any good. I only updated two columns and there are so many more. Maybe we can somehow forcefully change the encoding of our tables?

Additionally, only a very small amount of users is experiencing this issue, and we will be limiting the rest for no reason.

PeachesMLG (approved):
Looks good, didnt know if you were happy with it or not though. So merge it whenever ur ready

PR #619, opened 2022-02-17
Fix Big LAG when spam next page visit gui !

DoctaEnkoda:
Occurs when retrieving the UUID from the Mojang API.
Too many requests can crash the server!
Even the Bukkit.getOfflinePlayer is not enough because it is very slow!

Need : 
- https://github.com/Iridium-Development/IridiumCore/pull/31 (Support UUID Item)
- https://github.com/Iridium-Development/IridiumCore/pull/32 (Check UUID Exist in Item)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -24,7 +24,7 @@ repositories {
 dependencies {
     // Dependencies that we want to shade in
     implementation("org.jetbrains:annotations:22.0.0")
-    implementation("com.iridium:IridiumCore:1.5.1")
+    implementation("com.iridium:IridiumCore:1.5.4")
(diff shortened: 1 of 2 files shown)


dlsf (changes requested):
Rest looks good

dlsf (approved):

PeachesMLG:
Gonna be honest, I dont really like this, but only because it removes a bit of configurability.

What if somone doesnt want to have a player head in the /is members GUI?

Or is this chance so low that das will tell me to shut up and merge it? :thinking:

DoctaEnkoda:
Close because conflict

PR #621, opened 2022-02-19
Change how users bypass island privacy

dlsf:


DoctaEnkoda:
I am totally against this PR, because bypass gives access to the island (break, take) and for example our mods need this permission for anti afk rounds, giving them such bypass permission is destructive of a server .

PeachesMLG:
> 

Ah yeah, thats why we ended up having this. Should we do bypass or has permission then @dlsf?

dlsf:
That's why I hate responding to Github issues via email, you never know if it goes through... Sent this a few days ago:
> Oh, good point. But do you have anything against leaving the permission but adding the bypass as an alternative?

PR #622, opened 2022-02-20
Remove Duplication Code

DoctaEnkoda:
https://github.com/Iridium-Development/IridiumCore/blob/5a71acf539e0a2871691811eb2cdf3908d2c745b/plugin/src/main/java/com/iridium/iridiumcore/IridiumCore.java#L85


PeachesMLG:
Did you test this? IIRC We added the GUI interface into IridiumCore after we did one in IridiumSkyblock.

Although they are both called GUI IridiumCore's GUI and IridiumSkyblock's GUI are different objects

DoctaEnkoda:
ah yes my bad

PR #638, opened 2022-03-21
WorldEdit incompatibility check

boiscljo:
Verify if WorldEdit is incompatible with the current Minecraft version and switch back to default integration with a warning if that is the case.

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
@@ -30,10 +30,10 @@ public class SchematicManager {
         File parent = new File(IridiumSkyblock.getInstance().getDataFolder(), "schematics");
         SchematicPaster schematicPaster = worldEdit || fawe ? new WorldEdit() : new Schematic();
         
-        if ((worldEdit||fawe) && !WorldEdit.isWorking())
+        if ((worldEdit || fawe) && !WorldEdit.isWorking())
         {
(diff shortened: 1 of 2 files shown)


src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
```diff
@@ -28,7 +28,13 @@
 
     public SchematicManager() {
         File parent = new File(IridiumSkyblock.getInstance().getDataFolder(), "schematics");
-        this.schematicPaster = worldEdit || fawe ? new WorldEdit() : new Schematic();
+        boolean isWorldEditBroken = (worldEdit||fawe)?!WorldEdit.isWorking():false;
+        if (isWorldEditBroken)
```
PeachesMLG, line 32: I think this should be changed to
```suggestion
        if (!WorldEdit.isWorking() && (worldEdit||fawe))
```

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
boiscljo: Actually it would need to be the other way around (Otherwise it would launch a class not found in the isWorking if worldedit or fawe isn't there, I'll edit it tho.

dlsf (changes requested):
I'll request the actual changes once you implemented Peaches suggestions

dlsf (approved):

PR #674, opened 2022-06-22
Fix/cooldown clearing

Serega007RU:
This fixes a memory leak, you can find out more here https://discord.com/channels/543944856498864169/670664382140645418/987754464465260697


dlsf:
1) The build is broken because you removed an import
2) I don't think creating a new thread pool for every CooldownProvider is a good idea
3) I don't think scheduling a task for each cooldown is a good idea

PR #694, opened 2023-03-02
Gradle 8 + clean up

JT122406:
Cleans some stuff up, Updates Gradle to 8.0.1 and then fixed build issues because of it.  Update Dependencies and Repositories.  Move annotations to compileOnly


JT122406:
I realize I need to target other branch

PR #708, opened 2023-05-03
Update README.md

tarna:
fix a minor spelling error


dlsf (approved):
That README is very poorly phrased, but nice catch!

PR #718, opened 2023-07-06
improved paster from 3.2.x

boiscljo:
Reimplement FAWE support and paster improvement for lag issues from latest 3.2.x versions


src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
```diff
@@ -43,11 +43,16 @@ public Configuration() {
     public String worldName = "IridiumSkyblock";
     public String spawnWorldName = "world";
     public String islandTitleTop = "&9%island_name%";
-    public String islandTitleBottom = "&7%island_description%";
+    public String islandTitleBottom = "&7%island_description%";   
+    public String paster = "worldedit";
```
PeachesMLG, line 47: would it be worth considering making this an enum?

PR #724, opened 2023-08-07
Added regenInventoryClear

sh0inx:
Added the config options clearInventoryOnRegen & clearEnderChestOnRegen to allow servers to clear player inventories when regenerating islands.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,7 +5,7 @@ plugins {
 }
 
 group = "com.iridium"
-version = "4.0.4"
+version = "4.0.5"
 description = "IridiumSkyblock"
 
 repositories {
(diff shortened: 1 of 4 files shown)


CuBeR116:
Yes, the desired functionality. Players can endlessly recreate islands for themselves and transfer resources to their friends

sh0inx:
yeah... that's why the PR was made...
do you have anything you'd like to contribute to the PR?

CuBeR116:
Yes, it would also be forbidden to open a bank in other worlds of your choice.
If there are several worlds on the server, then players from different worlds can put crystals, money, experience in their bank.

Of course, if there is only SkyBlock on the server, then there will be no such problems.

sh0inx:
actually it would be preferable for you to just join our discord server

PR #736, opened 2023-08-21
impl. Island Delete Event

lackoSK:


src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -208,6 +209,8 @@ public void deleteTeam(Island island, User user) {
         IridiumSkyblock.getInstance().getDatabaseManager().getIslandTableManager().delete(island);
 
         getMembersOnIsland(island).forEach(member -> PlayerUtils.teleportSpawn(member.getPlayer()));
+
+        Bukkit.getPluginManager().callEvent(new IslandDeleteEvent(island, user));
```
boiscljo, line 213: Should check the return value and not call the delete methods if the event is cancelled (hence calling it before like 209 with a if statement about the cancelled state

boiscljo (approved):

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -202,6 +203,13 @@ private void deleteIslandBlocks(Island island, World world, int y, CompletableFu
 
     @Override
     public void deleteTeam(Island island, User user) {
+        final IslandDeleteEvent islandDeleteEvent = new IslandDeleteEvent(island, user);
+        Bukkit.getPluginManager().callEvent(islandDeleteEvent);
+
+        if (islandDeleteEvent.isCancelled()) {
+            return;
```
PeachesMLG, line 210: might be worth modifying this method to return a boolean.

In Teams, https://github.com/Iridium-Development/IridiumTeams/blob/master/src/main/java/com/iridium/iridiumteams/commands/DeleteCommand.java#L62 we will still send the deleted message. if this returns a boolean we can optionally send that.

Would require changing the https://github.com/Iridium-Development/IridiumTeams/blob/master/src/main/java/com/iridium/iridiumteams/managers/TeamManager.java#L89 Interface.

Feel free to open a PR in teams yourself, or ill do it myself over the weekend

PR #741, opened 2023-09-23
Ocean generator

PeachesMLG:


src/main/java/com/iridium/iridiumskyblock/configs/Generators.java
PeachesMLG, line 1: Are Biomes neccisary here? they will get overridden when you create an island, so will just be more confusing to people configuring no?

src/main/java/com/iridium/iridiumskyblock/generators/VoidGenerator.java
```diff
@@ -15,7 +18,13 @@ public class VoidGenerator extends ChunkGenerator {
 
     @Override
     public @NotNull ChunkData generateChunkData(@NotNull World world, @NotNull Random random, int chunkX, int chunkZ, @NotNull BiomeGrid biomeGrid) {
-        return createChunkData(world);
+        final ChunkData chunkData = createChunkData(world);
+        for (int x = 0; x < 16; x++) {
+            for (int z = 0; z < 16; z++) {
+                biomeGrid.setBiome(x, z, Objects.requireNonNull(getSkyblockGenerator(world.getEnvironment()).biome.getBiome()));
+            }
+        }
```
PeachesMLG, line 26: Possibly redundent

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -283,6 +284,69 @@ private void deleteIslandBlocks(Island island, World world, int y, CompletableFu
...
+
+                WorldCreator worldCreator = new WorldCreator(worldName)
+                        .generator(IridiumSkyblock.getInstance().getDefaultWorldGenerator(world.getName(), null))
+                        .environment(world.getEnvironment())
+                        .seed(world.getSeed());
+
+                worldCreator.createWorld();
+            }
+
+            World regenWorld = Bukkit.getWorld(worldName);
```
PeachesMLG, line 327: Is it creating this world in a single tick? does that cause any lag issues?

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -283,6 +284,69 @@ private void deleteIslandBlocks(Island island, World world, int y, CompletableFu
...
+                completableFuture.complete(null);
+            } else {
+                if (delay < 1) {
+                    regenerateTerrain(island, world, y - 1, completableFuture, delay);
+                } else {
+                    Bukkit.getScheduler().runTaskLater(IridiumSkyblock.getInstance(), () -> regenerateTerrain(island, world, y - 1, completableFuture, delay), delay);
+                }
+            }
+        }
+    }
```
PeachesMLG, line 348: Potentially we can utilise the Paster class, which should already have this code in

sh0inx: i can look into this, but i separated it into its own method because it will only run this if it detects we're using a terrain generator

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
sh0inx: uh probably, yeah

src/main/java/com/iridium/iridiumskyblock/configs/Generators.java
sh0inx: i added them for consistency, they do affect things spawning outside of an island for terrain generation purposes, but you're right

src/main/java/com/iridium/iridiumskyblock/managers/GeneratorType.java
```diff
@@ -0,0 +1,11 @@
+package com.iridium.iridiumskyblock.managers;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@Getter
+@AllArgsConstructor
+public enum GeneratorType {
+    VOID(false), OCEAN(true), VANILLA(true);
```
dlsf, line 9: This is not really how you format an enum with arguments

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -53,13 +52,37 @@ public IslandManager() {
...
+            for(int i=0; i<files.length; i++) {
+                if(files[i].isDirectory()) {
+                    deleteWorld(files[i]);
+                } else {
+                    files[i].delete();
+                }
+            }
+        }
+        return(path.delete());
+    }
```
dlsf, line 67: I can't believe there isn't really a better method for this in Java 8 without libraries... I suppose you could do this:

```suggestion
    public boolean deleteWorld(File path) {
        Files.walk(pathToBeDeleted)
            .sorted(Comparator.reverseOrder())
            .map(Path::toFile)
            .forEach(File::delete);
    }
```

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -53,13 +52,37 @@ public IslandManager() {
...
+
+    // When creating the cached world, we specify its name with a path to place it in the data folder.
+    // For some silly reason, using getDataFolder() results in an illegal character error.
+    // Same with using File.separator.
+    public String getCacheWorldName(World world) {
+        return "plugins/iridiumskyblock" + "/" + "regenWorlds" + "/" + world.getName() + "_regen";
```
dlsf, line 73: Use `JavaPlugin#getDataFolder()` instead

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -84,6 +107,30 @@ public void createWorld(World.Environment environment, String name) {
...
+                    .environment(world.getEnvironment())
+                    .seed(world.getSeed());
+
+            worldCreator.createWorld();
+        }
+
+        if (Bukkit.getWorld(getCacheWorldName(world)).getSeed() != Bukkit.getWorld(world.getName()).getSeed()) {
+            File cacheWorld = Bukkit.getWorld(getCacheWorldName(world)).getWorldFolder();
+            Bukkit.unloadWorld(getCacheWorldName(world), false);
+            deleteWorld(cacheWorld);
```
dlsf, line 129: What happens if the server crashes or the code above throws an exception? Maybe we should cleanup cached worlds on server startup?

dlsf (changes requested):

src/main/java/com/iridium/iridiumskyblock/managers/GeneratorType.java
dlsf:
```suggestion
    VOID(false), 
    OCEAN(true),
    VANILLA(true);
    
```

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
dlsf: Just read the comment, lmao. What exactly is the stack trace?

dlsf: I know that Paper is considering moving configs to a different folder in the long term, so something like this is less than optimal

sh0inx:
```console
[14:09:06 WARN]: [IridiumSkyblock] Task #38 for IridiumSkyblock v4.0.9-OCEAN generated an exception
java.lang.IllegalArgumentException: Invalid key. Must be [a-z0-9/._-]:
```
i would definitely prefer using
```java
IridiumSkyblock.getInstance().getDataFolder().getName() + File.separator + "regenWorlds" + File.separator + world.getName() + "_regen"
```
but uh yeah its bein dumb

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
sh0inx: we need to have the cached world for terrain generation, generating them on startup for every server start seems a bit unnecessary if the world is already there and matches our criteria

src/main/java/com/iridium/iridiumskyblock/generators/VoidGenerator.java
sh0inx: i dont remember why this was here, but i remember it being important, so if it breaks you can just revert the change ig

sh0inx:
We need to implement one of the following:

A)  Block populators for versions below 1.17.1
B)  Block populators for ALL versions, stop using the decorate function included in 1.17.1
C) Version check to ignore world type in versions below 1.17.1
D) Version check to ignore the decorate function in versions below 1.17.1

ideally, we have the generators work the same across all versions that IridiumSkyblock is compatible with, so I think either option A or B would be the preferred choice (i hate the concept too)

PR #747, opened 2023-10-02
Added payPerRegen setting

sh0inx:
- added `payPerRegen` setting to allow servers to charge players to regenerate their island
- includes a message sent to players of the island when purchased with `%schematic% placeholder to add schematic name to message
- missing schematic lore

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
+++ b/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
@@ -47,7 +47,7 @@ public class RegenCommand extends Command<Island, User> {
 
         Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(schematic.get());
 
-        if(schematicConfig.regenCost.money != 0 && !schematicConfig.regenCost.bankItems.isEmpty()) {
+        if(schematicConfig.regenCost.money != 0 || !schematicConfig.regenCost.bankItems.isEmpty()) {
             if(!IridiumSkyblock.getInstance().getSchematicManager().buy(player, schematicConfig)) {
                 return false;
             }
--- a/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
+++ b/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
@@ -48,8 +48,7 @@ public class RegenCommand extends Command<Island, User> {
         Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(schematic.get());
 
         if(IridiumSkyblock.getInstance().getConfiguration().payPerRegen) {
-            boolean purchased = IridiumSkyblock.getInstance().getSchematicManager().buy(player, schematicConfig);
-            if(!purchased) {
+            if(!IridiumSkyblock.getInstance().getSchematicManager().buy(player, schematicConfig)) {
                 return false;
             }
         }
--- a/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
+++ b/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
@@ -48,7 +48,7 @@ public class RegenCommand extends Command<Island, User> {
         Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(schematic.get());
 
         if(schematicConfig.regenCost.money != 0 && !schematicConfig.regenCost.bankItems.isEmpty()) {
-            if(!IridiumSkyblock.getInstance().getSchematicManager().processTransaction(player, schematicConfig)) {
+            if(!IridiumSkyblock.getInstance().getSchematicManager().buy(player, schematicConfig)) {
                 return false;
             }
         }
--- a/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
+++ b/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
@@ -47,8 +47,8 @@ public class RegenCommand extends Command<Island, User> {
 
         Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(schematic.get());
 
-        if(schematicConfig.regenCost.money == 0 && schematicConfig.regenCost.bankItems.isEmpty()) {
-            if(!IridiumSkyblock.getInstance().getSchematicManager().buy(player, schematicConfig)) {
+        if(!(schematicConfig.regenCost.money == 0 && schematicConfig.regenCost.bankItems.isEmpty())) {
+            if(!IridiumSkyblock.getInstance().getSchematicManager().processTransaction(player, schematicConfig)) {
                 return false;
             }
         }
--- a/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
+++ b/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
@@ -47,6 +47,13 @@ public class RegenCommand extends Command<Island, User> {
 
         Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(schematic.get());
 
+        if(IridiumSkyblock.getInstance().getConfiguration().payPerRegen) {
+            boolean purchased = IridiumSkyblock.getInstance().getSchematicManager().buy(player, schematicConfig);
+            if(!purchased) {
+                return false;
+            }
+        }
+
         IridiumSkyblock.getInstance().getIslandManager().getTeamMembers(island).stream()
                 .map(User::getPlayer)
                 .filter(Objects::nonNull)
--- a/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
+++ b/src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
@@ -47,7 +47,7 @@ public class RegenCommand extends Command<Island, User> {
 
         Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(schematic.get());
 
-        if(IridiumSkyblock.getInstance().getConfiguration().payPerRegen) {
+        if(schematicConfig.regenCost.money == 0 && schematicConfig.regenCost.bankItems.isEmpty()) {
             if(!IridiumSkyblock.getInstance().getSchematicManager().buy(player, schematicConfig)) {
                 return false;
             }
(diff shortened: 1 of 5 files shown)


src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
```diff
@@ -47,6 +47,13 @@ public boolean execute(User user, Island island, String[] args, IridiumTeams<Isl
 
         Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(schematic.get());
 
+        if(IridiumSkyblock.getInstance().getConfiguration().payPerRegen) {
+            boolean purchased = IridiumSkyblock.getInstance().getSchematicManager().buy(player, schematicConfig);
+            if(!purchased) {
```
PeachesMLG, line 52:
```suggestion
            if(!IridiumSkyblock.getInstance().getSchematicManager().buy(player, schematicConfig)) {
```

src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
```diff
@@ -53,6 +53,7 @@ public Configuration() {
     public boolean removeIslandBlocksOnDelete = false;
     public boolean clearInventoryOnRegen = false;
     public boolean clearEnderChestOnRegen = false;
+    public boolean payPerRegen = false;
```
PeachesMLG, line 56: Can we not remove this config option, and instead disable it if the schematic cost <= 0?

src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
```diff
@@ -115,4 +125,78 @@ private CompletableFuture<Void> pasteSchematic(Island island, Schematics.Schemat
...
+        }
+
+        return moneyCost == 0 || economy != null && economy.getBalance(player) >= schematic.regenCost.money;
+    }
+
+    private void purchase(Player player, Schematics.SchematicConfig schematic) {
+        IridiumSkyblock.getInstance().getEconomy().withdrawPlayer(player, schematic.regenCost.money);
+
+        for (String bankItem : schematic.regenCost.bankItems.keySet()) {
+            setBankBalance(player, bankItem, getBankBalance(player, bankItem) - round(schematic.regenCost.bankItems.get(bankItem), 2));
```
PeachesMLG, line 186: not a fan of this since you retrieve the object twice, its probs not a huge difference, but technically you are wasting performance by doing the same operation twice to recieve the same object

sh0inx: i think im stupid, what object

src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
sh0inx: aye aye, captain

src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
```diff
@@ -47,7 +47,7 @@ public boolean execute(User user, Island island, String[] args, IridiumTeams<Isl
 
         Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(schematic.get());
 
-        if(IridiumSkyblock.getInstance().getConfiguration().payPerRegen) {
+        if(schematicConfig.regenCost.money == 0 && schematicConfig.regenCost.bankItems.isEmpty()) {
```
PeachesMLG, line 50: This would be if its not 0, and its not empty right?

src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
PeachesMLG: They both call the [IslandManager#getTeamBank](https://github.com/Iridium-Development/IridiumSkyblock/blob/6292d20aa2111933b7970d2e739db43befb45af4/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java#L330) method, whilst this shouldnt be too bad performance wise, depending on how many islands you have, it may have to search through a few thousand entries, so we should try to minimise this where possible

sh0inx: ok this might be more of a coding question then; if we set bank as a variable, which is how i would go about solving this, does the code run that once to determine the value and keep that until we update it ourselves, or does calling the variable call the method every time we call it?

src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
sh0inx: uh yeah, that should be right
this is why i tend to have too many parentheses, i prefer to write them like math equations 
ex:
```java
if((schematicConfig.regenCost.money == 0) && (schematicConfig.regenCost.bankItems.isEmpty()))
```

src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
```diff
@@ -115,4 +125,78 @@ private CompletableFuture<Void> pasteSchematic(Island island, Schematics.Schemat
...
+        Optional<Island> island = IridiumSkyblock.getInstance().getTeamManager().getTeamViaID(user.getTeamID());
+
+        if(!island.isPresent()){
+            //NO ISLAND
+            return false;
+        }
+
+        TeamBank bank;
+
+        double bankBalance;
```
PeachesMLG, line 140:
```suggestion
```

src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
```diff
@@ -115,4 +125,78 @@ private CompletableFuture<Void> pasteSchematic(Island island, Schematics.Schemat
...
+        List<Placeholder> bankPlaceholders;
+        double moneyCost = schematic.regenCost.money;
+        Economy economy = IridiumSkyblock.getInstance().getEconomy();
+
+        if(moneyCost == 0 || economy != null && economy.getBalance(player) >= moneyCost) {
+            //INVALID TRANSACTION
+            return false;
+        }
```
PeachesMLG, line 148: I think this needs to be flipped?

In the origional code this was if the player can afford to buy the item. So we need to flip it and return false if they cant afford to buy it?

```suggestion
        if(moneyCost != 0 && economy != null && economy.getBalance(player) < moneyCost) {
            //INVALID TRANSACTION
            return false;
        }
```

PeachesMLG, line 148: I think this needs to be flipped?

In the origional code this was if the player can afford to buy the item. So we need to flip it and return false if they cant afford to buy it?

```suggestion
        if(moneyCost != 0 && economy != null && economy.getBalance(player) < moneyCost) {
            //Cannot Afford
            return false;
        }
```

src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
```diff
@@ -47,6 +47,12 @@ public boolean execute(User user, Island island, String[] args, IridiumTeams<Isl
 
         Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(schematic.get());
 
+        if(!(schematicConfig.regenCost.money == 0 && schematicConfig.regenCost.bankItems.isEmpty())) {
```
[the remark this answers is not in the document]
PeachesMLG: Im also wondering if you can use the schematicConfig.canPurchase() method?

src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
```diff
@@ -115,4 +125,78 @@ private CompletableFuture<Void> pasteSchematic(Island island, Schematics.Schemat
...
+
+            double cost = round(schematic.regenCost.bankItems.get(schematicBankItem), 2);
+
+            if (bank.getNumber() < cost) {
+                player.sendMessage(StringUtils.color(IridiumSkyblock.getInstance().getMessages().cannotAfford
+                        .replace("%prefix%", IridiumSkyblock.getInstance().getConfiguration().prefix)));
+                IridiumSkyblock.getInstance().getSchematics().failSound.play(player);
+
+                return false;
+            }
```
PeachesMLG, line 163: This needs to be moved outside the for loop, so we check if they can afford it before we start taking stuff away

src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
```diff
@@ -115,4 +125,78 @@ private CompletableFuture<Void> pasteSchematic(Island island, Schematics.Schemat
...
+                    .collect(Collectors.toList());
+
+            player.sendMessage(StringUtils.color(StringUtils.processMultiplePlaceholders(IridiumSkyblock.getInstance().getMessages().paidForRegen
+                            .replace("%prefix%", IridiumSkyblock.getInstance().getConfiguration().prefix)
+                            .replace("%player%", player.getName())
+                            .replace("%schematic%", StringUtils.color(schematic.item.displayName))
+                            .replace("%bank_cost%", formatPrice(cost))
+                            .replace ("%vault_cost%", formatPrice(moneyCost)),
+                    bankPlaceholders)
+            ));
```
PeachesMLG, line 178: This also needs to be outside the for loop otherwise we will send the player a message for each BankItem

src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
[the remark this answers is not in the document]
PeachesMLG: Also note, this change may look redundant, but there is a difference in the logic.

The previous change made it so you had to have BOTH a moneyCost and a bankitem, but the new chance will change it to needing to have alteast one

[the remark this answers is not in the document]
sh0inx: 🤔 probably

src/main/java/com/iridium/iridiumskyblock/managers/SchematicManager.java
sh0inx: yeah, i wanted to add bank items to the message, tho im not sure how to add each and every bank item and then just the one vault eco without separating the message (rn they cant see they're paying both)

src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
```diff
@@ -47,6 +47,12 @@ public boolean execute(User user, Island island, String[] args, IridiumTeams<Isl
 
         Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(schematic.get());
 
+        if(schematicConfig.regenCost.money != 0 && !schematicConfig.regenCost.bankItems.isEmpty()) {
```
PeachesMLG, line 50:
```suggestion
        if(schematicConfig.regenCost.money != 0 || !schematicConfig.regenCost.bankItems.isEmpty()) {
```

I think this should be or no?

PeachesMLG: for example if the vault money cost is 0, but I say it costs 5 island crystals, this should still be valid?

Or If I said it was 10,000 but 0 bank items, thats also valid, no?

PR #751, opened 2023-10-06
Update IslandManager.java

sh0inx:
updated the island manager to automagically skip the createGUI for the user if there is only 1 schematic to pick from

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
@@ -134,31 +134,27 @@ public class IslandManager extends TeamManager<Island, User> {
         return IridiumSkyblock.getInstance().getDatabaseManager().getIslandTableManager().getEntries();
     }
 
-    @Override
-    public CompletableFuture<Island> createTeam(@NotNull Player owner, String name) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -137,13 +137,28 @@ public List<Island> getTeams() {
...
+            } else {
+                schematic = schematicNameCompletableFuture.join();
+            }
+
+            String finalSchematic = schematic;
+            if (finalSchematic == null) return null;
 
             User user = IridiumSkyblock.getInstance().getUserManager().getUser(owner);
-            Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(schematic);
+            Schematics.SchematicConfig schematicConfig = IridiumSkyblock.getInstance().getSchematics().schematics.get(finalSchematic);
```
PeachesMLG, line 161: Would moving this to a new method make more sence?

private CompletableFuture<String> getSchematic(Player owner);

sh0inx: it would probably enhance code readability, but as far as functionality goes it seems to be working fine
if we want to we can do that

PR #753, opened 2023-10-08
killed the dragon

sh0inx:
used NBTAPI to edit level.dat values and prevent the dragon or bedrock portal from spawning because im cool like that (should fix 749)

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/IridiumSkyblock.java
+++ b/src/main/java/com/iridium/iridiumskyblock/IridiumSkyblock.java
@@ -89,13 +89,9 @@ public class IridiumSkyblock extends IridiumTeams<Island, User> {
 
         this.teamManager = new IslandManager();
 
-        try {
-            this.teamManager.createWorld(World.Environment.NORMAL, configuration.worldName);
-            this.teamManager.createWorld(World.Environment.NETHER, configuration.worldName + "_nether");
-            this.teamManager.createWorld(World.Environment.THE_END, configuration.worldName + "_the_end");
-        } catch (IOException e) {
-            throw new RuntimeException(e);
-        }
+        this.teamManager.createWorld(World.Environment.NORMAL, configuration.worldName);
+        this.teamManager.createWorld(World.Environment.NETHER, configuration.worldName + "_nether");
+        this.teamManager.createWorld(World.Environment.THE_END, configuration.worldName + "_the_end");
 
         this.schematicManager = new SchematicManager();
         this.userManager = new UserManager();
--- a/src/main/java/com/iridium/iridiumskyblock/IridiumSkyblock.java
+++ b/src/main/java/com/iridium/iridiumskyblock/IridiumSkyblock.java
@@ -89,9 +89,13 @@ public class IridiumSkyblock extends IridiumTeams<Island, User> {
 
         this.teamManager = new IslandManager();
 
-        this.teamManager.createWorld(World.Environment.NORMAL, configuration.worldName);
-        this.teamManager.createWorld(World.Environment.NETHER, configuration.worldName + "_nether");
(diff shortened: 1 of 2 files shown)


src/main/java/com/iridium/iridiumskyblock/IridiumSkyblock.java
```diff
@@ -89,9 +89,13 @@ public void onEnable() {
...
-        this.teamManager.createWorld(World.Environment.NORMAL, configuration.worldName);
-        this.teamManager.createWorld(World.Environment.NETHER, configuration.worldName + "_nether");
-        this.teamManager.createWorld(World.Environment.THE_END, configuration.worldName + "_the_end");
+        try {
+            this.teamManager.createWorld(World.Environment.NORMAL, configuration.worldName);
+            this.teamManager.createWorld(World.Environment.NETHER, configuration.worldName + "_nether");
+            this.teamManager.createWorld(World.Environment.THE_END, configuration.worldName + "_the_end");
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
```
PeachesMLG, line 98: Why was this done?

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -49,12 +52,44 @@ public IslandManager() {
...
+
+            if(PreviouslyKilled == (byte) 0) {
+                compound.setByte("PreviouslyKilled", (byte) 1);
+            }
+            if(DragonKilled == 0) {
+                compound.setByte("DragonKilled", (byte) 1);
+            }
+            if(NeedsStateScanning == 1) {
+                compound.setByte("NeedsStateScanning", (byte) 0);
+            }
```
PeachesMLG, line 87: Can we not change this to setBoolean? iirc it just wraps setByte and sets it to 0/1

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -49,12 +52,44 @@ public IslandManager() {
...
+
+            worldFile.save();
+
+            Bukkit.createWorld(worldCreator);
```
PeachesMLG, line 91: Are we saving this before its created? it should create the world first no?

PeachesMLG (commented):
I dont particually like this method, but it might be the best option

src/main/java/com/iridium/iridiumskyblock/IridiumSkyblock.java
sh0inx: IDE wouldn't let me compile without it

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
PeachesMLG: Also, you can probabbly remove the if statements no?

sh0inx: i haven't tried that actually, i'll give it a shot and see what we end up with (this was done based on me looking at the NBTdata directly so)

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
sh0inx: no because it has to save ``level.dat`` before loading the world

src/main/java/com/iridium/iridiumskyblock/IridiumSkyblock.java
PeachesMLG: hmm ill look into it later, seems weird

GNUGradyn (approved):

PR #758, opened 2023-10-13
Bank items craft fix

Faun471:


Faun471:
Moved to iridiumteams.

PR #761, opened 2023-11-05
Added whiteListedWorlds config option

sh0inx:
see [teams pr](https://github.com/Iridium-Development/IridiumTeams/pull/31)

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
+++ b/src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
@@ -8,6 +8,7 @@ import com.iridium.iridiumcore.dependencies.xseries.XSound;
 import com.iridium.iridiumteams.Reward;
 import org.bukkit.World;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
```diff
@@ -35,6 +36,11 @@ public Configuration() {
                         "&9&l[!] &9Left click to redeem"
                 )), Collections.emptyList(), 0, new ImmutableMap.Builder<String, Double>().put("Crystals", 10.00).build(), 2000, 0, XSound.ENTITY_PLAYER_LEVELUP))
                 .build();
+
+        this.whitelistedWorlds = new ArrayList<>();
+            whitelistedWorlds.add(worldName);
+            whitelistedWorlds.add(worldName + "_nether");
+            whitelistedWorlds.add(worldName + "_the_end");
```
PeachesMLG, line 43: Can we not do Arrays.asList() ?

src/main/java/com/iridium/iridiumskyblock/listeners/PlayerPortalListener.java
```diff
@@ -16,6 +16,19 @@ public class PlayerPortalListener implements Listener {
...
+            if(event.getPlayer().getWorld().getName().equalsIgnoreCase(world)) {
+                setCancelled = false;
+                break;
+            }
+        }
+
+        if(setCancelled) {
+            return;
+        }
+
```
PeachesMLG, line 31: This shouldnt be needed since we already do a check for if an island is present, which already does a World check

src/main/java/com/iridium/iridiumskyblock/listeners/PlayerInteractListener.java
```diff
@@ -36,6 +36,18 @@ public void onClick(PlayerInteractEvent event) {
...
+            if(player.getWorld().getName().equalsIgnoreCase(world)) {
+                setCancelled = false;
+                break;
+            }
+        }
+
+        if(setCancelled) {
+            return;
+        }
+
```
PeachesMLG, line 50: This shouldnt be needed since we already do a check for if an island is present, which already does a World check

PR #765, opened 2023-11-11
Actually Island Create on Join

sh0inx:
i am bad at github sorry

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerJoinListener.java
+++ b/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerJoinListener.java
@@ -3,19 +3,17 @@ package com.iridium.iridiumskyblock.listeners;
 import com.iridium.iridiumskyblock.IridiumSkyblock;
 import org.bukkit.Bukkit;
 import org.bukkit.event.EventHandler;
+import org.bukkit.event.EventPriority;
 import org.bukkit.event.Listener;
 import org.bukkit.event.player.PlayerJoinEvent;
 
 public class PlayerJoinListener implements Listener {
 
-    @EventHandler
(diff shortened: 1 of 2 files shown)


src/main/java/com/iridium/iridiumskyblock/listeners/PlayerJoinListener.java
```diff
@@ -1,15 +1,19 @@
...
+import org.bukkit.event.EventPriority;
 import org.bukkit.event.Listener;
 import org.bukkit.event.player.PlayerJoinEvent;
 
 public class PlayerJoinListener implements Listener {
 
-    @EventHandler
+    @EventHandler (priority = EventPriority.MONITOR)
     public void onPlayerJoin(PlayerJoinEvent event) {
+        if(IridiumSkyblock.getInstance().getConfiguration().islandCreateOnJoin && !event.getPlayer().hasPlayedBefore())
```
PeachesMLG, line 14: do we wana check if they have an island isntead of seeing if they played before?

I guess the difference is if they create an island, delete the island and then relog, should it make a new one for them?

sh0inx: the idea is to have the server put them in the create menu as soon as they join the server for the first time, we could add a config option for if the server wants to bring this prompt up after island deletion and rejoin but I feel that it may be a lil bit unnecessary considering they already have experience with the server/plugin

sh0inx: because the create menu is skipped on a single schematic, I imagine this will also just create an island for them if they don't have one, which they won't because it's their first join

Issue #768, opened 2023-12-07
[ BUG ] Console error on new island


SkyDude0:
**Describe the bug**
Console Error https://pastebin.com/nxJ6CGyv

**To Reproduce**
1. Create new island
2. Check console for the error

**Expected behavior**
No error

**Server and Plugin Version**
SERVER: 1.18.2

PL: 4.0.8

sh0inx:
for some reason the plugin cant load the world
strange, provide your entire log via https://mclo.gs please

sh0inx:
is ``spawn1`` your IridiumSkyblock world? could be related to world name, try renaming it to IridiumSkyblock

sh0inx:
okay then where is your skyblock world? because it only loaded the one world, and iridiumSkyblock cant really do anything without its world loaded

sh0inx:
havent had the time to check on this

sh0inx:
fixed in [IS 774](https://github.com/Iridium-Development/IridiumSkyblock/pull/774)

PR #774, opened 2023-12-12
fixed null world on biome set

sh0inx:
see [IS 768](https://github.com/Iridium-Development/IridiumSkyblock/issues/768)


src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -231,9 +227,23 @@ public CompletableFuture<Void> generateIsland(Island island, Schematics.Schemati
...
+                    case NETHER:
+                        setIslandBiome(island, XBiome.matchXBiome(schematicConfig.nether.biome));
+                        break;
+                    case THE_END:
+                        setIslandBiome(island, XBiome.matchXBiome(schematicConfig.end.biome));
+                        break;
+                    default:
+                        setIslandBiome(island, XBiome.matchXBiome(schematicConfig.overworld.biome));
+                    }
+            }
```
PeachesMLG, line 246: Can this section not stay how it is? since we already handle if the world is enabled in setIslandBiome no?

I think the previous code is alot cleaner

sh0inx: ill do some testing, i also found it a bit redundant but i wanted to make sure we got rid of the problem

this still doesn't help anything if the worlds are enabled/disabled during server runtime - this is a separate issue tho

PeachesMLG (approved):

PR #787, opened 2024-01-07
Prevent obsidian platform generate in The End

### LINKED ISSUE #779: Obsidian platform abnormally generate
**Describe the bug**
Each time the player enters the end, the end will generate a new obsidian platform at a place with a lower part of the platform that was generated before, and remove all the surrounding blocks (Vanilla behavior)

**To Reproduce**
1. Build a end portal.
2. Jump in
3. A new obsidian platform generated (suppose its location is 0 90 0)
4. Execute `/is home` to return overworld
5. Jump in the portal again
6. A **NEW** obsidian platform generated and we will found its location is 0 89 0.
7. Repeat 4-6, and the platform will be lower and lower

**Expected behavior**
Obsidian platform shound not generate. Or it shound not remove any blocks.

**Server and Plugin Version**
`This server is running Purpur version git-Purpur-1417 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 69c7c31 on ver/1.17.1)`
`Plugin Version: 4.0.8`

**Screenshots**

First join the end, I am in y=87

Rejoin the end, I am in y=86

**Logs**
No errors or wanings displayed in console.

**Additional context**
The config is default and only Vault installed.

DreamVoid:
Fix #779 
It works well.
Also fixed player not being in the correct location after teleported. (Old behavior players will be teleported inside the block)

`setCanCreatePortal` may be incompatible with 1.15-, but it seems possible to abandon 1.15-?

image

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerPortalListener.java
+++ b/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerPortalListener.java
@@ -62,6 +62,7 @@ public class PlayerPortalListener implements Listener {
                     event.setCancelled(true);
                     return;
                 }
+                event.setCanCreatePortal(false);
                 event.setTo(location);
             }
         });
--- a/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerPortalListener.java
+++ b/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerPortalListener.java
@@ -1,5 +1,6 @@
 package com.iridium.iridiumskyblock.listeners;
 
+import com.iridium.iridiumcore.dependencies.xseries.XMaterial;
 import com.iridium.iridiumcore.utils.StringUtils;
 import com.iridium.iridiumskyblock.IridiumSkyblock;
 import com.iridium.iridiumskyblock.utils.LocationUtils;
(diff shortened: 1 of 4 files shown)


src/main/java/com/iridium/iridiumskyblock/listeners/PlayerPortalListener.java
```diff
@@ -62,6 +62,8 @@ public void onPlayerPortal(PlayerPortalEvent event) {
                     event.setCancelled(true);
                     return;
                 }
+                location.setY(location.getY() + 1);
+                event.setCanCreatePortal(false);
```
PeachesMLG, line 66:
```suggestion
                if(XMaterial.supports(16)) event.setCanCreatePortal(false);
```
We could do this to try maintain 1.15- This wont fix the obsidian platform for 1.15, but atleast it will still be somewhat supported

DreamVoid: Origin code even doesn't support 1.16- ¯\_(ツ)_/¯
Because `java.lang.NoSuchFieldError: NETHER_WASTES`

PeachesMLG (approved):

Issue #802, opened 2024-02-01
Regen island doesn't remove entities.


Micosedeunha:
**Describe the bug**
Using /is regen does not remove entities from the island.

**To Reproduce**
1. Create an island
2. Spawn some zombies and place armor stands and item frames
3. Use /regen island and select the same schematic you had before (just for better understand)
4. look around, zombies and armor stands does not disappear

**Expected behavior**
Regenerate completely the island, remove all item frames, armor stands and mobs

**Server and Plugin Version**
Current: git-Purpur-2095 (MC: 1.20.2)*
![image](https://github.com/Iridium-Development/IridiumSkyblock/assets/154468237/9586e34f-42c9-4042-bda5-763bc2362602)

**Screenshots**
before regen
![2024-02-01_13 35 09](https://github.com/Iridium-Development/IridiumSkyblock/assets/154468237/0b2d2132-70c4-4937-bc5b-b14f6c6c4772)
after regen (I translated to portuguese, but in chat says that island has regenerated)
![image](https://github.com/Iridium-Development/IridiumSkyblock/assets/154468237/f4d8bb87-3794-4e78-8cb4-eb3be07f3ae5)

PR #803, opened 2024-02-02
added clearEntities method

Faun471:
This should remove all non-player entities upon island regen


Faun471:
This should fix #802

src/main/java/com/iridium/iridiumskyblock/commands/RegenCommand.java
```diff
@@ -80,6 +81,7 @@ public boolean execute(User user, Island island, String[] args, IridiumTeams<Isl
         ));
 
         IridiumSkyblock.getInstance().getIslandManager().clearTeamInventory(island);
+        IridiumSkyblock.getInstance().getIslandManager().clearEntities(island);
```
PeachesMLG, line 84: This should  probabbly be moved inside the generateIsland method

PeachesMLG (approved):

PR #804, opened 2024-02-11
made the default island name to be displayed configurable

Faun471:
I have not tested this...

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/database/Island.java
+++ b/src/main/java/com/iridium/iridiumskyblock/database/Island.java
@@ -137,6 +137,6 @@ public class Island extends Team {
                 .findFirst()
                 .map(User::getName)
                 .orElse("N/A");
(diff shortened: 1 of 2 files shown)


src/main/java/com/iridium/iridiumskyblock/database/Island.java
```diff
@@ -137,6 +137,6 @@ public void setColor(Color color) {
                 .findFirst()
                 .map(User::getName)
                 .orElse("N/A");
-        return ownerName + "'s Island";
+        return IridiumSkyblock.getInstance().getConfiguration().defaultIslandName.replace("%player%", ownerName);
```
PeachesMLG, line 140:
```suggestion
        return IridiumSkyblock.getInstance().getConfiguration().defaultIslandName.replace("%owner%", ownerName);
```

PeachesMLG (approved):

PR #811, opened 2024-02-20
Fix Api bug and add standard getOwner/getMembers

boiscljo:
Fixes the static block initializing in API and adds missing getOwner/getMembers

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
+++ b/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
@@ -44,7 +44,9 @@ public class IridiumSkyblockAPI {
      * @since 3.0.0
      */
     public static IridiumSkyblockAPI getInstance() {
-        if(instance==null && IridiumSkyblock.getInstance()!=null) instance = new IridiumSkyblockAPI(IridiumSkyblock.getInstance());
+        if(instance == null && IridiumSkyblock.getInstance() != null) {
+            instance = new IridiumSkyblockAPI(IridiumSkyblock.getInstance());
+        }
         return instance;
     }
 
--- a/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
+++ b/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
@@ -43,7 +43,7 @@ public class IridiumSkyblockAPI {
      * @return the instance of this api
      * @since 3.0.0
      */
-    public static @NotNull IridiumSkyblockAPI getInstance() {
+    public static IridiumSkyblockAPI getInstance() {
         if(instance==null && IridiumSkyblock.getInstance()!=null) instance = new IridiumSkyblockAPI(IridiumSkyblock.getInstance());
         return instance;
     }
--- a/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
+++ b/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
@@ -43,7 +43,7 @@ public class IridiumSkyblockAPI {
      * @return the instance of this api
      * @since 3.0.0
      */
-    public static IridiumSkyblockAPI getInstance() {
+    public static @Nullable IridiumSkyblockAPI getInstance() {
         if(instance == null && IridiumSkyblock.getInstance() != null) {
             instance = new IridiumSkyblockAPI(IridiumSkyblock.getInstance());
         }
--- a/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
+++ b/src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
@@ -24,12 +24,8 @@ import java.util.Optional;
  */
 public class IridiumSkyblockAPI {
 
-    private static final IridiumSkyblockAPI instance;
+    private static IridiumSkyblockAPI instance;
     private final IridiumSkyblock iridiumSkyblock;
-
-    static {
-        instance = new IridiumSkyblockAPI(IridiumSkyblock.getInstance());
-    }
 
     /**
      * Constructor for api initialization.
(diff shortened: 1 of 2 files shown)


src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
```diff
@@ -48,6 +44,7 @@ private IridiumSkyblockAPI(@NotNull IridiumSkyblock iridiumSkyblock) {
      * @since 3.0.0
      */
     public static @NotNull IridiumSkyblockAPI getInstance() {
```
PeachesMLG, line 46: Remove the NotNull annotation since it can be nullable now, probabbly add some sort of comment saying how its null if the plugin hasnt loaded yet?

src/main/java/com/iridium/iridiumskyblock/database/Island.java
```diff
@@ -132,11 +134,22 @@ public void setColor(Color color) {
     @Override
     public @NotNull String getName() {
         if (super.getName() != null) return super.getName();
-        String ownerName = IridiumSkyblock.getInstance().getTeamManager().getTeamMembers(this).stream()
+        String ownerName = getMembers().stream()
                 .filter(user -> user.getUserRank() == Rank.OWNER.getId())
                 .findFirst()
```
PeachesMLG, line 139:
```suggestion
        String ownerName = GetOwner()
```

src/main/java/com/iridium/iridiumskyblock/database/Island.java
```diff
@@ -132,11 +134,22 @@ public void setColor(Color color) {
...
                 .map(User::getName)
                 .orElse("N/A");
         return IridiumSkyblock.getInstance().getConfiguration().defaultIslandName.replace("%owner%", ownerName);
     }
+
+    public List<User> getMembers()
+    {
+        return IridiumSkyblock.getInstance().getTeamManager().getTeamMembers(this);
+    }
+    public Optional<User> getOwner()
```
PeachesMLG, line 149: Can we add a space in between this please

src/main/java/com/iridium/iridiumskyblock/database/Island.java
```diff
@@ -132,11 +134,21 @@ public void setColor(Color color) {
...
+    {
+        return IridiumSkyblock.getInstance().getTeamManager().getTeamMembers(this);
+    }
+
+    public Optional<User> getOwner()
+    {
+        return getMembers().stream()
+                .filter(user -> user.getUserRank() == Rank.OWNER.getId())
+                .findFirst();
+    }
```
dlsf, line 153: Really don't like how inconsistent this is, why suddenly use the C# bracket style

src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
```diff
@@ -47,7 +43,8 @@ private IridiumSkyblockAPI(@NotNull IridiumSkyblock iridiumSkyblock) {
      * @return the instance of this api
      * @since 3.0.0
      */
-    public static @NotNull IridiumSkyblockAPI getInstance() {
+    public static IridiumSkyblockAPI getInstance() {
+        if(instance==null && IridiumSkyblock.getInstance()!=null) instance = new IridiumSkyblockAPI(IridiumSkyblock.getInstance());
```
dlsf, line 47: I think it would be better if you used a multiline if statement here for better readability

```suggestion
        if(instance == null && IridiumSkyblock.getInstance() != null) {
            instance = new IridiumSkyblockAPI(IridiumSkyblock.getInstance());
        }
```

I also think the concerns about this potentially returning null weren't addressed?

dlsf (changes requested):

src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
boiscljo: The comment above mentions it might return null during initialization.

dlsf: Ah, it was displaying the wrong Javadoc. How about an additional @Nullable for consistency with the other nullable API methods, though? It would also be more accessible

boiscljo: Can you send it as suggestions? Because I'm on movile and can't add it.

src/main/java/com/iridium/iridiumskyblock/api/IridiumSkyblockAPI.java
```diff
@@ -47,7 +43,10 @@ private IridiumSkyblockAPI(@NotNull IridiumSkyblock iridiumSkyblock) {
      * @return the instance of this api
      * @since 3.0.0
      */
-    public static @NotNull IridiumSkyblockAPI getInstance() {
+    public static IridiumSkyblockAPI getInstance() {
```
dlsf, line 46: Because I can't comment on outdated code:
```suggestion
    public static @Nullable IridiumSkyblockAPI getInstance() {
```

src/main/java/com/iridium/iridiumskyblock/database/Island.java
```diff
@@ -132,11 +134,19 @@ public void setColor(Color color) {
     @Override
     public @NotNull String getName() {
         if (super.getName() != null) return super.getName();
-        String ownerName = IridiumSkyblock.getInstance().getTeamManager().getTeamMembers(this).stream()
-                .filter(user -> user.getUserRank() == Rank.OWNER.getId())
-                .findFirst()
+        String ownerName = getOwner()
                 .map(User::getName)
                 .orElse("N/A");
```
dlsf, line 139: Not sure about 4.0, but in 3.0 this text was a config option. Do we still have it?

boiscljo: I have no clue, but it's out of scope of this PR since I didn't add it. If it's a config peaches or someone else should fix it

boiscljo: For reference N/A is used in 11 places in V4... should have a separe PR to remove the N/A strings

dlsf (approved):

PR #817, opened 2024-02-29
Created the ClearData command

sh0inx:
- self explanatory, but requires a change to the SQL manager to fix an error


sh0inx:
Somewhere along the way, a bug was introduced that caused the team to be deleted regardless of data type being cleared (should probably fix that...)

PeachesMLG (approved):

Issue #49, opened 2021-04-25
Unable to insert schematic data


dlsf:
When using MySQL as the SQL driver, I receive the following error message: https://gist.github.com/dlsf/2177ba637ae0e369a78c610290e44467

PR #818, opened 2024-03-01
Added Island Level Configuration

sh0inx:
- the other side of [IridiumTeams #49](https://github.com/Iridium-Development/IridiumTeams/pull/49)


src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
```diff
@@ -56,6 +56,7 @@ public Configuration() {
     public boolean clearEnderChestOnRegen = false;
     public boolean allowPvPOnIslands = false;
     public boolean islandCreateOnJoin = false;
+    public boolean islandCreationCost = false;
```
PeachesMLG, line 59: I feel like this is going to confuse things even more. I understand there are alot of people asking this in the Discord, but if we want to remove the creation cost, I feel setting the default cost to 0 for everything is a better aproach

src/main/java/com/iridium/iridiumskyblock/gui/BiomeCategoryGUI.java
```diff
@@ -61,7 +62,7 @@ public void addContent(Inventory inventory) {
 
     private List<Placeholder> getBiomeLorePlaceholders(Biomes.BiomeItem item) {
         List<Placeholder> placeholders = new ArrayList<>(Arrays.asList(
-                new Placeholder("amount", IridiumSkyblock.getInstance().getBiomeManager().formatPrice(item.defaultAmount)),
+                new Placeholder("minLevel", String.valueOf(item.minLevel)),
```
PeachesMLG, line 65: Do we not need the Amount placeholder anymore?

src/main/java/com/iridium/iridiumskyblock/managers/BiomeManager.java
```diff
@@ -78,14 +75,40 @@ private void setBankBalance(Player player, String bankItem, double amount) {
     }
 
     private boolean canPurchase(Player player, Biomes.BiomeItem biomeItem) {
+
+        if(biomeItem.minLevel != 1) {
```
PeachesMLG, line 79:
```suggestion
        if(biomeItem.minLevel > 1) {
```

src/main/java/com/iridium/iridiumskyblock/gui/BiomeCategoryGUI.java
sh0inx: yes, this allows us to fetch the price of the schematic from the config directly without having to set the lore manually every time

src/main/java/com/iridium/iridiumskyblock/configs/Configuration.java
sh0inx: this was more for the fact that we could allow pricing for creation *and* regen to be different (as in, when you create the island, it doesn't cost anything, but when you regen, it does) - we could change it to be free but honestly i dont care about those people lol

PeachesMLG (approved):

PR #827, opened 2024-03-27
Automatic releases to Modrinth and Hangar

sh0inx:
- added publish to Hangar, Modrinth, and Github
- added new check for incrementing beta versions


PeachesMLG:
This file is starting to get very big, wondering if we should move it into a publish file or something?

Also potentially running this when we create a github release: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release would be better than a commit

GNUGradyn (approved):

PeachesMLG (approved):

PR #840, opened 2024-05-07
Implemented Stacker Support

PeachesMLG:


src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -473,14 +473,24 @@ public CompletableFuture<Void> recalculateTeam(Island island) {
                         for (int y = 0; y <= maxy; y++) {
                             if (island.isInIsland(x + (chunkSnapshot.getX() * 16), z + (chunkSnapshot.getZ() * 16))) {
                                 XMaterial material = XMaterial.matchXMaterial(chunkSnapshot.getBlockType(x, y, z));
-                                teamBlocks.put(material, teamBlocks.getOrDefault(material, 0) + 1);
+                                Block block = chunk.getBlock(x, y, z);
+                                int amount = IridiumSkyblock.getInstance().getSupportManager().getStackerSupport().stream()
+                                        .mapToInt(stackerSupport -> stackerSupport.getStackAmount(block))
+                                        .max()
+                                        .orElse(1);
+                                teamBlocks.put(material, teamBlocks.getOrDefault(material, 0) + amount);
```
PeachesMLG, line 481: Need to think about this since we are doing this in an async thread

PR #842, opened 2024-05-07
you can respawn on your island now, or whatever

sh0inx:
also rewrote the player utils class to use the spawn support from teams, its pretty cool

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerRespawnEventListener.java
+++ b/src/main/java/com/iridium/iridiumskyblock/listeners/PlayerRespawnEventListener.java
@@ -1,6 +1,7 @@
 package com.iridium.iridiumskyblock.listeners;
 
 import com.iridium.iridiumskyblock.IridiumSkyblock;
+import com.iridium.iridiumskyblock.database.Island;
 import com.iridium.iridiumskyblock.utils.PlayerUtils;
 import lombok.AllArgsConstructor;
 import org.bukkit.event.EventHandler;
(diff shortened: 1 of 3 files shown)


src/main/java/com/iridium/iridiumskyblock/listeners/PlayerRespawnEventListener.java
```diff
@@ -0,0 +1,20 @@
...
+import org.bukkit.event.player.PlayerRespawnEvent;
+
+@AllArgsConstructor
+public class PlayerRespawnEventListener implements Listener {
+
+    @EventHandler(priority = EventPriority.HIGHEST)
+    public void onPlayerRespawnEvent(PlayerRespawnEvent event) {
+
+        if(!IridiumSkyblock.getInstance().getConfiguration().spawnOnIsland) { return; }
+        if(!event.isBedSpawn() && !event.isAnchorSpawn()) event.setRespawnLocation(PlayerUtils.getSpawn(event.getPlayer()));
```
PeachesMLG, line 18: Is it worth checking if the current spawn is inside the Island?

If we set our bed spawn at another players island, would we expect to respawn in somone elses island or at our island home?

src/main/java/com/iridium/iridiumskyblock/utils/PlayerUtils.java
```diff
@@ -1,26 +1,40 @@
...
+    }
+
+    public static Location getSpawn(Player player) {
+
+        if(IridiumSkyblock.getInstance().getConfiguration().spawnOnIsland) {
+            Optional<Island> island = IridiumSkyblock.getInstance().getTeamManager().getTeamViaNameOrPlayer(player.getName());
+            if(island.isPresent()) {
+                return island.get().getHome();
+            }
         }
```
PeachesMLG, line 27: iirc, this method is used to find a place to teleport a player when their island gets deleted. so not sure if this will break stuff

sh0inx: since we're using an optional here, if it does not return an island (which it shouldnt since the team was deleted), it should teleport the player to the spawn point

src/main/java/com/iridium/iridiumskyblock/listeners/PlayerRespawnEventListener.java
sh0inx: 🤔 im not 100% sure tbh, because you'd expect that someone would respawn at their bed regardless of where that bed is, since that is vanilla behavior
we can add a check to the interact permission to prevent players from setting their spawn on other player islands if this is a concern

src/main/java/com/iridium/iridiumskyblock/utils/PlayerUtils.java
PeachesMLG: It depends if we delete the island then teleport them, or teleport them and then delete the island... lemme check in code

sh0inx: kk

src/main/java/com/iridium/iridiumskyblock/listeners/PlayerRespawnEventListener.java
PeachesMLG: I think a check IridiumSkyblock.getInstance().getIslandManager().canVisit should be added

PeachesMLG (approved):

PR #853, opened 2024-05-17
Switched from NbtItem to Nbt.modify

sh0inx:
see https://github.com/Iridium-Development/IridiumCore/pull/99

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
@@ -702,8 +702,9 @@ public class IslandManager extends TeamManager<Island, User> {
         if (itemStack == null || itemStack.getType() == Material.AIR) return 0;
 
         return NBT.get(itemStack, readableItemNBT -> {
-            if(readableItemNBT.getCompound("iridiumskyblock").hasTag("islandCrystals"))
+            if(readableItemNBT.getCompound("iridiumskyblock").hasTag("islandCrystals")) {
                 return readableItemNBT.getCompound("iridiumskyblock").getInteger("islandCrystals");
+            }
             return 0;
         });
     }
--- a/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
@@ -701,15 +701,11 @@ public class IslandManager extends TeamManager<Island, User> {
     public int getIslandCrystals(ItemStack itemStack) {
         if (itemStack == null || itemStack.getType() == Material.AIR) return 0;
 
-        int amount = 0;
-
-        if (NBT.get(itemStack, readableItemNBT -> {
-            return readableItemNBT.getCompound("iridiumskyblock").hasTag("islandCrystals");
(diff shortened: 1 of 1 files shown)


src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -690,19 +690,26 @@ public ItemStack getIslandCrystal(int amount) {
...
-        NBTCompound nbtCompound = new NBTItem(itemStack).getOrCreateCompound("iridiumskyblock");
-        if (nbtCompound.hasKey("islandCrystals")) {
-            return nbtCompound.getInteger("islandCrystals");
-        }
-        return 0;
+
+        int amount = 0;
+
+        if (NBT.get(itemStack, readableItemNBT -> {
+            return readableItemNBT.getCompound("iridiumskyblock").hasTag("islandCrystals");
```
boiscljo, line 707: can`t you combine both? would make more sense

sh0inx: yeah probably but i couldnt figure it out without the IDE complaining

sh0inx: feel free to post a suggestion

boiscljo: return NBT.get(itemStack, readableItemNBT -> {
           if(readableItemNBT.getCompound("iridiumskyblock").hasTag("islandCrystals"))
            return readableItemNBT.getCompound("iridiumskyblock").getInteger("islandCrystals");
            return 0;
        });

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -690,19 +690,22 @@ public ItemStack getIslandCrystal(int amount) {
...
-        NBTCompound nbtCompound = new NBTItem(itemStack).getOrCreateCompound("iridiumskyblock");
-        if (nbtCompound.hasKey("islandCrystals")) {
-            return nbtCompound.getInteger("islandCrystals");
-        }
-        return 0;
+
+        return NBT.get(itemStack, readableItemNBT -> {
+            if(readableItemNBT.getCompound("iridiumskyblock").hasTag("islandCrystals"))
+                return readableItemNBT.getCompound("iridiumskyblock").getInteger("islandCrystals");
+            return 0;
```
PeachesMLG, line 707: I think in general you should add brackets onto if statements unless its on the same line


e.g. the only place where its ok to not have brackets is something like

```
if (thing) return
```

if you did

``` if(thing)
return;
```

This shouldnt be allowed (its very easy for bugs to appear since it gets confusing whats in and whats not inside the if statement otherwise)

PeachesMLG (changes requested):

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
sh0inx: roger doger

PeachesMLG (approved):

PR #868, opened 2024-06-15
Added slot removal option to island menu

sh0inx:
honestly cant believe it was a one line fix smh

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/managers/CommandManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/CommandManager.java
@@ -33,6 +33,8 @@ public class CommandManager extends com.iridium.iridiumteams.managers.CommandMan
             Player player = (Player) commandSender;
             User user = IridiumSkyblock.getInstance().getUserManager().getUser(player);
             if (IridiumSkyblock.getInstance().getTeamManager().getTeamViaID(user.getTeamID()).isPresent()) {
+
+                IridiumSkyblock.getInstance().getInventories().islandMenu.items.values().removeIf(item -> item.slot == -1);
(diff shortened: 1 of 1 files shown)


src/main/java/com/iridium/iridiumskyblock/managers/CommandManager.java
```diff
@@ -33,6 +33,8 @@ public void noArgsDefault(@NotNull CommandSender commandSender) {
             Player player = (Player) commandSender;
             User user = IridiumSkyblock.getInstance().getUserManager().getUser(player);
             if (IridiumSkyblock.getInstance().getTeamManager().getTeamViaID(user.getTeamID()).isPresent()) {
+
+                IridiumSkyblock.getInstance().getInventories().islandMenu.items.values().removeIf(item -> item.slot == -1);
```
PeachesMLG, line 37: So this will remove the thing from the config, if the slot is -1.

Alternitavly you can also add an if statement here https://github.com/Iridium-Development/IridiumTeams/blob/master/src/main/java/com/iridium/iridiumteams/gui/InventoryConfigGUI.java#L36
So we just dont add it to the GUI if the slot is < 0.


I dont really know what the better solution is here, I guess it really just depends on if we want it to be deleted from the config or not

PeachesMLG (approved):

src/main/java/com/iridium/iridiumskyblock/managers/CommandManager.java
PeachesMLG: Im not really fussed either way, so ive approved this and ill let you decide on which one you want to go for

sh0inx: i would rather use [this](https://github.com/Iridium-Development/IridiumTeams/pull/75), you can go ahead and close this pr in favor of teams

PR #876, opened 2024-07-17
Generate plugin.yml dynamically

dlsf:
No particular reason, but it's cooler, less hacky & shorter


PeachesMLG:
Idk about this, feels like there is no benifit to this, and it would potentially just make things more confusing in the future.


I can definatly see a scenario where in a months time I need to add a soft dependancy and ive completly forgotten all about this, and cant remember where this stuff is managed.


Or if a new contributor wants to add something, I can see them getting confused with not having a plugin.yml

@sh0inx Any thoughts from you?

sh0inx:
i like the idea of managing everything we need from a single file, but i will parrot the initial confusion from a potential contributor as a potential issue, simply because most plugins are expected to operate in a specific way.
does this work with paper/folia?

PeachesMLG (approved):
Spoke about this on discord, and people seemed in favour on this, We will proceed with this method for now, and in the future if it causes issues we can always revert this change

PR #879, opened 2024-07-19
Fix inaccurate block stacks

Faun471:
Stacked Blocks always have one more than the actual amount, since we count the placed block as well as the amount of blocks inside the "stack"

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
+++ b/src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
@@ -516,26 +516,8 @@ public class IslandManager extends TeamManager<Island, User> {
         return CompletableFuture.runAsync(() -> {
             List<Chunk> chunks = getIslandChunks(island).join();
             for (Chunk chunk : chunks) {
-                ChunkSnapshot chunkSnapshot = chunk.getChunkSnapshot(true, false, false);
-                for (int x = 0; x < 16; x++) {
-                    for (int z = 0; z < 16; z++) {
-                        final int maxy = chunkSnapshot.getHighestBlockYAt(x, z);
-                        for (int y = 0; y <= maxy; y++) {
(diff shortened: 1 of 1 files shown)


Faun471:
I made a lot of changes with how the code looks; I can change it back to how it looked before for consistency idk

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -549,6 +531,42 @@ public CompletableFuture<Void> recalculateTeam(Island island) {
...
+                    if (island.isInIsland(x + (chunkSnapshot.getX() * 16), z + (chunkSnapshot.getZ() * 16))) {
+                        processBlock(x, y, z, chunk, chunkSnapshot, teamBlocks);
+                    }
+                }
+            }
+        }
+    }
+
+    private void processBlock(int x, int y, int z, @NotNull Chunk chunk, @NotNull ChunkSnapshot chunkSnapshot, @NotNull Map<XMaterial, Integer> teamBlocks) {
+        Block block = chunk.getBlock(x, y, z);
```
PeachesMLG, line 549: Is this thread safe? Remember this is running async

src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -549,6 +531,42 @@ public CompletableFuture<Void> recalculateTeam(Island island) {
...
+        XMaterial material = XMaterial.matchXMaterial(chunkSnapshot.getBlockType(x, y, z));
+        int amount = isBlockStacked(block) ? getStackAmount(block) : 1;
+        teamBlocks.merge(material, amount, Integer::sum);
+    }
+
+    public boolean isBlockStacked(Block block) {
```
PeachesMLG, line 555: Is this needed? getStackAmount defaults to 1 if its not present anyway?

PeachesMLG (changes requested):

PR #930, opened 2024-12-28
Fixed Issues with recieving multiple rewards

PeachesMLG:


src/main/java/com/iridium/iridiumskyblock/managers/IslandManager.java
```diff
@@ -779,7 +779,9 @@ public List<TeamReward> getTeamRewards(Island island) {
 
     @Override
     public void addTeamReward(TeamReward teamReward) {
-        IridiumSkyblock.getInstance().getDatabaseManager().getTeamRewardsTableManager().addEntry(teamReward);
+        CompletableFuture.runAsync(() -> IridiumSkyblock.getInstance().getDatabaseManager().getTeamRewardsTableManager().save(teamReward)).thenRun(() ->
+                IridiumSkyblock.getInstance().getDatabaseManager().getTeamRewardsTableManager().addEntry(teamReward)
+        );
```
PeachesMLG, line 784: This saves the record to the database first, which will generate the Id field.

What was happening was the id wasnt set, so all defaulted to 0, and here https://github.com/Iridium-Development/IridiumSkyblock/blob/master/src/main/java/com/iridium/iridiumskyblock/managers/DatabaseManager.java#L85 You can see we are keying by the id.

So when we save, it actually does a save/update and will replace if the key is the same, but since we havent saved to the database yet, the key (Id) isnt set, so is 0, which means that it will override the other values

PeachesMLG:
This also patches the issue with being able to infinitly redeem rewards.

How we store the records is in a hashmap of key/value, as I mentioned above, we set the rewards key to the database id.

However, this will also be set to 0, which isnt itself an issue.

However, if you wait 5 minutes for the database to save, it will assign the id 0 to the next sequentual value, however it wont update the hashmap key, which means when we try to delete the value, we try to delete key, lets say 5, however no key exists under 5 in the hashmap (because its under 0)

This should also fix that

PR #1129, opened 2026-08-25
Feature/deepslate generator

rrvk:
With this pull request i am added Add deepslate ore support for cobblestone generators below Y=0

Behavior
- Above Y=0: unchanged (stone/cobblestone + regular ores)
- Below Y=0: deepslate + deepslate ore variants
- Basalt generators: unchanged (still uses nether ore table)
- Existing configs without deepslateOres: gracefully falls back to normal ores

I have tested this on a 26.1.2 paper server

Ps its my first pull request here let me know if there is anything else i need to do :)

### DIFF
--- a/src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
+++ b/src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
@@ -16,6 +16,7 @@ public class BlockFormListener implements Listener {
 
     private final Map<Integer, RandomAccessList<XMaterial>> normalOreLevels = new HashMap<>();
     private final Map<Integer, RandomAccessList<XMaterial>> netherOreLevels = new HashMap<>();
+    private final Map<Integer, RandomAccessList<XMaterial>> deepslateOreLevels = new HashMap<>();
 
     private final List<XMaterial> generatorMaterials = Arrays.asList(XMaterial.STONE, XMaterial.COBBLESTONE, XMaterial.BASALT);
 
@@ -23,6 +24,9 @@ public class BlockFormListener implements Listener {
         for (Map.Entry<Integer, GeneratorEnhancementData> oreUpgrade : IridiumSkyblock.getInstance().getEnhancements().generatorEnhancement.levels.entrySet()) {
             normalOreLevels.put(oreUpgrade.getKey(), new RandomAccessList<>(oreUpgrade.getValue().ores));
             netherOreLevels.put(oreUpgrade.getKey(), new RandomAccessList<>(oreUpgrade.getValue().netherOres));
+            if (oreUpgrade.getValue().deepslateOres != null) {
+                deepslateOreLevels.put(oreUpgrade.getKey(), new RandomAccessList<>(oreUpgrade.getValue().deepslateOres));
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


sh0inx:
this actually looks pretty good, great work on your first pr!

 i want to ask though, do you think that it might be a good idea to provide this as a config option? I can imagine this might mess with existing islands and settings, since any generators that someone has below y = 0 will now generate with the default settings you've supplied here.
 
 did you test how the config migrates to using your new generator options btw? it should be a seamless transition, meaning that the config is recognized and the deepslate generation values are populated automagically to an existing configuration.

rrvk:
Thanks!

It kinda has a feature toggle. Because if you run this version with a existing enchantments.yml this will be added:  deepslateOres: null
(Example)
```
 levels:
    0:
      minLevel: 5
      money: 10000
      bankCosts:
        Crystals: 5.0
      ores:
        COBBLESTONE: 3
        COAL_ORE: 1
      netherOres:
        BASALT: 1
      deepslateOres: null
```
      
That way no deepslate stuff will be generated. 

If you start a brand new server with this version the default options will be generated

sh0inx:
wouldnt it being null cause issues tho?

rrvk:
https://github.com/Iridium-Development/IridiumSkyblock/pull/1129/changes/0bc6d7e5767ad0d977759d5d1533c40d28868791#diff-361be43f3977b7bd90326b539a98146aa1ce6ed00ab85046336a1fe3429f8a7fR27

That will check on null values

src/main/java/com/iridium/iridiumskyblock/listeners/BlockFormListener.java
```diff
@@ -16,13 +16,17 @@ public class BlockFormListener implements Listener {
...
 
     private final List<XMaterial> generatorMaterials = Arrays.asList(XMaterial.STONE, XMaterial.COBBLESTONE, XMaterial.BASALT);
 
     public BlockFormListener() {
         for (Map.Entry<Integer, GeneratorEnhancementData> oreUpgrade : IridiumSkyblock.getInstance().getEnhancements().generatorEnhancement.levels.entrySet()) {
             normalOreLevels.put(oreUpgrade.getKey(), new RandomAccessList<>(oreUpgrade.getValue().ores));
             netherOreLevels.put(oreUpgrade.getKey(), new RandomAccessList<>(oreUpgrade.getValue().netherOres));
+            if (oreUpgrade.getValue().deepslateOres != null) {
+                deepslateOreLevels.put(oreUpgrade.getKey(), new RandomAccessList<>(oreUpgrade.getValue().deepslateOres));
+            }
```
PeachesMLG, line 29: on the else here, I would just put the normal ores inside deepslate ores.

So if deepslate is null it fallsback to the normal ores

rrvk: I can do that but that seems a little strange because they are not the deepslateOres but the normal ores.
What i have done is this `deepslateOreLevels.getOrDefault(upgradeLevel, normalOreLevels.get(upgradeLevel)` (line 42) Because that way if there is nothing in deepslateMap if will just use the normalOre map

But I can change it if you want it?

PeachesMLG: ok yeah, I see that now. I just want it so as said above if someone upgrades to this the deepslate will be null in the config. if this happens there shouldnt be a change in functionality, so it should use the normalOres.

PeachesMLG:
Nice one, this looks good thanks

